PR #746, opened 2014-12-10
Trigger DOM event of type 'INPUT' after changing value of INPUT element

frolix:
When using native keyboard to send keys, then DOM event of type 'input' is triggered automatically by platform.

Opposite for AndroidWebElement, when setting value for INPUT element, then no DOM event of type 'input' is triggered.

This fix will trigger DOM event type 'input' after setting value of web element 'INPUT'.

AndroidWebElement is activated when Selenium session request has property nativeEvents set to false.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,13 @@ Maven Artifacts
 
 You can find the current version in Maven central and the latest snapshot version here: [https://oss.sonatype.org/content/repositories/snapshots/](https://oss.sonatype.org/content/repositories/snapshots/)
 
+Building
+--------
+1. Install Cordova framework artifact to local .m2 repo by executing:
+```
+$ third-party/cordova-3.7.0/mvnInstall.sh
+```
+2. Run ```mvn clean compile package```
 
 License
 -----------
--- a/selendroid-server/src/main/java/io/selendroid/server/model/AndroidWebElement.java
+++ b/selendroid-server/src/main/java/io/selendroid/server/model/AndroidWebElement.java
@@ -335,7 +335,11 @@ public class AndroidWebElement implements AndroidElement {
     parameter.put(this);
     parameter.put(getText() + sb.toString());
 
-    driver.executeScript("arguments[0].value = arguments[1]", parameter, ke);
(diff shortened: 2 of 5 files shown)


DominikDary:
Hi, thanks for your pull request. 
Can you please add a test for this changed behavior?

frolix:
@DominikDary I will add the test!

DominikDary:
@frolix any update on writing the test?

frolix:
Hi @DominikDary, please, can you help on what should I do to execute unit tests in selendroid-test-app module? mvn test won't help..
Maybe a doc or smth?

frolix:
@DominikDary I've managed to execute the tests via ```runEnd2EndTests.sh```, though there were some issues (not sure if I did everything the right way).

1. when test, then Selendroid fails to push ```extension.dex``` to emulator (```io.selendroid.support.BaseAndroidTest#getDefaultCapabilities```), I had always to do it manually via
```adb -s emulator-5560 push target/test-classes/extension.dex /storage/sdcard/extension.dex```
when emulator starts
2. for some reason ```runEnd2EndTests.sh``` at step "Running End-to-End Tests" runs all tests twice.

Not sure if it is critical.

Unit test is created. Now there are two tests concerning ```sendKeys```:
* ```shouldTriggerInputEventWhenSendTextWithNativeKeyboard``` - to ensure that sending text via Native Keyboard causes ```input``` event to be triggered
* ```shouldTriggerInputEventWhenSendTextWithoutNativeKeyboard``` - to ensure that sending text via JavaScript is same as when with Native Keyboard (except ```keydown``` and ```keyup``` events, which are not much relevant in this case)

Is everything ok with the commits and tests?

frolix:
A simple selenium test can help to verify the behaviour with input event:
1. open google.com
2. find input element and call ```sendKeys``` method on it

expected: dropdown box with suggestions should appear

the expected behaviour should work same - with capability
```CapabilityType.HAS_NATIVE_EVENTS``` set to **true** (the default) as well as when set to **false**

DominikDary:
Awesome, thanks for the PR and the test!

PR #756, opened 2014-12-16
Grid auto re-register

frolix:
If a grid hub was restarted, then SelendroidStandaloneServer will re-register itself again.

Commit details:
  a task created for self registration - first check grid if the node already registered, and re-register if not.
  then the task is scheduled for execution (recurrence is configured via -registerCycle argument) 

A nice feature that I found useful when having a farm of selendroid-servers. :-)

### DIFF
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/SelendroidConfiguration.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/SelendroidConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2014 eBay Software Foundation and selendroid committers.
+ * Copyright 2012-2015 eBay Software Foundation and selendroid committers.
  * 
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -15,15 +15,12 @@ package io.selendroid.standalone;
 
 import com.beust.jcommander.JCommander;
 import com.beust.jcommander.Parameter;
-
 import io.selendroid.standalone.log.LogLevelConverter;
 import io.selendroid.standalone.log.LogLevelEnum;
-
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.apache.commons.lang3.builder.ToStringStyle;
 
-import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -98,6 +95,10 @@ public class SelendroidConfiguration {
 
   @Parameter(names = "-maxInstances", description = "Maximum number of instances that a grid hub can use at a time.")
   private int maxInstances = 5;
+
+  @Parameter(names = "-registerCycle", description = "How often in ms the node will try to register itself again" +
+          ".Allow to restart the hub without having to restart the nodes. 0 to disable auto register. Default 0.")
+  private long registerCycle = 0;
 
   @Parameter(names = "-noWebviewApp",
              description = "If you don't want selendroid to auto-extract and have 'AndroidDriver' (webview only app) available.")
@@ -369,4 +370,13 @@ public class SelendroidConfiguration {
   public boolean isGrid() {
     return !StringUtils.isBlank(getRegistrationUrl()) && !StringUtils.isBlank(getServerHost());
   }
+
+  public long getRegisterCycle() {
+    return registerCycle;
+  }
+
+  public void setRegisterCycle(long registerCycle) {
+    this.registerCycle = registerCycle;
+  }
+
 }
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/server/SelendroidStandaloneServer.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/server/SelendroidStandaloneServer.java
@@ -21,12 +21,10 @@ import io.selendroid.standalone.exceptions.AndroidSdkException;
 import io.selendroid.standalone.server.grid.SelfRegisteringRemote;
 import io.selendroid.standalone.server.model.SelendroidStandaloneDriver;
 
-import java.util.concurrent.TimeUnit;
+import java.util.Timer;
+import java.util.TimerTask;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-
-import io.selendroid.standalone.server.util.HttpClientUtil;
-import org.apache.commons.lang3.StringUtils;
 
 public class SelendroidStandaloneServer {
   private static final Logger log = Logger.getLogger(SelendroidStandaloneServer.class.getName());
@@ -67,14 +65,32 @@ public class SelendroidStandaloneServer {
 
   public void start() {
     webServer.start();
+
     if (config.isGrid()) {
+      selfRegisterInGrid();
+    }
+
+    log.info("Selendroid standalone server has been started on port: " + config.getPort());
+  }
+
+  private void selfRegisterInGrid() {
+    final SelfRegisteringRemote selfRegisteringRemote = new SelfRegisteringRemote(config, driver);
+
+    if (config.getRegisterCycle() > 0) {
+      log.info("Scheduling periodic task for self registration in GRID.");
+      new Timer().schedule(new TimerTask() {
+        @Override
+        public void run() {
+          selfRegisteringRemote.performRegistrationIfNotRegistered();
+        }
+      }, 0, config.getRegisterCycle());
+    } else {
       try {
-        new SelfRegisteringRemote(config, driver).performRegistration();
+        selfRegisteringRemote.performRegistration();
       } catch (Exception e) {
         log.log(Level.SEVERE, "Error registering selendroid into grid hub.", e);
       }
     }
-    log.info("Selendroid standalone server has been started on port: " + config.getPort());
   }
 
   public void stop() {
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/server/grid/SelfRegisteringRemote.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/server/grid/SelfRegisteringRemote.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2014 eBay Software Foundation and selendroid committers.
+ * Copyright 2012-2015 eBay Software Foundation and selendroid committers.
  * 
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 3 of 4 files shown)


selendroid-standalone/src/main/java/io/selendroid/standalone/SelendroidConfiguration.java
```diff
@@ -99,6 +99,10 @@
   @Parameter(names = "-maxInstances", description = "Maximum number of instances that a grid hub can use at a time.")
   private int maxInstances = 5;
 
+  @Parameter(names = "-registerCycle", description = "How often in ms the node will try to register itself again" +
```
asm89: Should there be a way to disable this behavior?

selendroid-standalone/src/main/java/io/selendroid/standalone/server/SelendroidStandaloneServer.java
```diff
@@ -68,11 +66,41 @@ protected SelendroidStandaloneDriver initializeSelendroidServer() throws Android
...
+          this.driver = driver;                // passing reference to driver and config
+          this.configuration = configuration;  // as, let's say, config might change during runtime... (device quantity)
+        }
+
+        @Override
+        public void run() {
+          try {
+            final SelfRegisteringRemote remote = new SelfRegisteringRemote(configuration, driver);
+
+            if (firstTime) {
```
asm89: no need for this if, you can always set `firstTime` to `false`. The checking happens in `performRegistrationIfNotRegistered()`.

selendroid-standalone/src/main/java/io/selendroid/standalone/server/grid/SelfRegisteringRemote.java
```diff
@@ -47,6 +51,12 @@ public SelfRegisteringRemote(SelendroidConfiguration config, SelendroidStandalon
     this.driver = driver;
   }
 
+  public void performRegistrationIfNotRegistered(final boolean firstTime) throws Exception {
+    if (firstTime || !isAlreadyRegistered()){
```
asm89: Maybe it would be better to make this class aware of whether this happens the first time or not (make `firstTime` a bool here).

selendroid-standalone/src/main/java/io/selendroid/standalone/server/grid/SelfRegisteringRemote.java
```diff
@@ -136,7 +146,7 @@ private JSONObject getConfiguration() throws JSONException {
       configuration.put("proxy", "org.openqa.grid.selenium.proxy.DefaultRemoteProxy");
     }
     configuration.put("role", "node");
-    configuration.put("registerCycle", 5000);
+    configuration.put("registerCycle", config.getRegisterCycle());
```
asm89: What does this value mean? You named the startup parameter the same, but the value now got changed from `5s` to `60s`..

selendroid-standalone/src/main/java/io/selendroid/standalone/server/grid/SelfRegisteringRemote.java
```diff
@@ -156,4 +166,40 @@ private JSONObject getConfiguration() throws JSONException {
     configuration.put("remoteHost", "http://" + config.getServerHost() + ":" + config.getPort());
     return configuration;
   }
+
+  private boolean isAlreadyRegistered() {
+    HttpClient client = HttpClientUtil.getHttpClient();
+
+    try {
+      String tmp = config.getRegistrationUrl().replace("/grid/register", "/grid/api/proxy"); //TODO find a better way
```
asm89: Can you fix this TODO before merging?

asm89:
I don't know too much about grid. It would be great if someone with more experience could review this as well. 

/cc @selendroid/contributers

koichirok:
Since Selendroid add/remove hardware device dynamically, you could not use same logic as Selenium Grid for "isAlreadyRegistered()" check.

selendroid-standalone/src/main/java/io/selendroid/standalone/server/grid/SelfRegisteringRemote.java
koichirok: I think 60s is quite unusable.

koichirok:
there seems to be possible race condition with small registerCycle.

selendroid-standalone/src/main/java/io/selendroid/standalone/SelendroidConfiguration.java
frolix: @asm89, thanks! Good point. 
Do you think that setting registerCycle to 0 would be sufficient to disable this re-register behavior? (just to not introduce any other flags)

selendroid-standalone/src/main/java/io/selendroid/standalone/server/grid/SelfRegisteringRemote.java
frolix: @asm89, the **registerCycle** value is time in milliseconds between recurrences of following operation:
 * check if already registered at hub. The check will return false when not registered. For example: a grid hub was restarted unexpectedly.
 * if the check above returned false, then perform registration at grid hub.

Originally in grid sources - [SelfRegisteringRemote.java](https://code.google.com/p/selenium/source/browse/java/server/src/org/openqa/grid/internal/utils/SelfRegisteringRemote.java), line 210:
`        Thread.sleep(registerCycleInterval);`

The default value for **registerCycleInterval** in code is set to **0** (i.e. recurrence is disabled)
[SelfRegisteringRemote.java](https://code.google.com/p/selenium/source/browse/java/server/src/org/openqa/grid/internal/utils/SelfRegisteringRemote.java), line 191:
`       final int registerCycleInterval = nodeConfig.getConfigAsInt(RegistrationRequest.REGISTER_CYCLE, 0);`


@asm89 and @koichirok, the value of 5000ms for **registerCycle** was set by me earlier, I've found it somewhere in sources as a default value.

But later experiments showed that when having a farm of over 100 of selendroid nodes with 4-6 emulators each will produce too much HTTP traffic just for the registration checks and result in flooding network and unnecessary load for grid hub.
So I've found a 60k milliseconds is a quite reasonable interval for checks. I think it could be set as high as equivalent of 5 minutes, because grid hub restarts happen not that often. But that's about personal choice. 

I'd like to suggest to set the default value to 0 as it will be inline with @asm89 point above about disabling the recurrence behavior and with original grid's sources [SelfRegisteringRemote.java](https://code.google.com/p/selenium/source/browse/java/server/src/org/openqa/grid/internal/utils/SelfRegisteringRemote.java), line 191.
`       final int registerCycleInterval = nodeConfig.getConfigAsInt(RegistrationRequest.REGISTER_CYCLE, 0);`

What is your opinion?

P.S.: original Selenium's references that might be of interest
[GridParameters.properties](https://code.google.com/p/selenium/source/browse/java/server/src/org/openqa/grid/common/defaults/GridParameters.properties) and [Grid2 wiki](https://code.google.com/p/selenium/wiki/Grid2#Optional_parameters)

Also, originally, in code design of [SelfRegisteringRemote.java](https://code.google.com/p/selenium/source/browse/java/server/src/org/openqa/grid/internal/utils/SelfRegisteringRemote.java) authors decided to use Thread.sleep in a while-true cycle to perform recurrence. I'd think that using Java API's  Timer and TimerTask would fit better.

selendroid-standalone/src/main/java/io/selendroid/standalone/server/grid/SelfRegisteringRemote.java
frolix: @asm, good idea! I'll move the bool to the SelfRegisteringRemote class.

selendroid-standalone/src/main/java/io/selendroid/standalone/server/SelendroidStandaloneServer.java
frolix: @asm89, the check for firstTime is a kind of '--force-register' mode. The filed use for this is - let's say, you would like to restart selendroid for some reason (upgrade or config changes - emulators/devices/other grid parameters). And selendroid restarted quite fast, so grid hub have not noticed that selendroid node was absent. Then without this '--force-register' mode, the selendroid node will not send new config to grid hub.

selendroid-standalone/src/main/java/io/selendroid/standalone/server/grid/SelfRegisteringRemote.java
frolix: Well, this is kinda for discussion. 
I've left this TODO here as currently there is no better way to do it. 
See, that selendroid's only knowledge of hub is **'-hub'** argument. Also, see in [SelendroidConfiguration.java](https://github.com/selendroid/selendroid/blob/master/selendroid-standalone/src/main/java/io/selendroid/standalone/SelendroidConfiguration.java) that a argument **-hub** which is actually not a hub, but a URL for registering (i.e. it is not a host name or port, but a string with URL - http://localhost:4444/grid/register or something). 
The best, if it would be hub-host and hub-port, as originally in grid sources, then there would no be this TODO. But I guess it's too late to change this.

At this point, perhaps a more correct way (not faster, but more correct), would be:
1. Take string value of **'-hub'** argument
2. Create a URL object of it
3. Take Scheme, host and port parts from URL object
4. Form a new URL with path **'/grid/api/proxy'**

If that looks like an OK refactoring, I'll apply the change.

frolix:
As @koichirok said: 
>Since Selendroid add/remove hardware device dynamically, you could not use same logic as Selenium Grid for "isAlreadyRegistered()" check.

Well, I'd say that actually you could. :-)
See that the reference to instance of SelendroidConfiguration is passed as a first parameter to SelfRegisteringRemote constructor.
So just plug/unplug your devices and call performRegistration method to notify grid hub.

(I actually use it for notifying grid hub if one of emulators crashed and that selendroid node has less free browsers while crashed emulator is rebuilding and restarting (this code is not published yet. not sure if it could be of use)).

koichirok:
@frolix wrote:
> As @koichirok said:
>
> > Since Selendroid add/remove hardware device dynamically, you could not use same logic as Selenium Grid for "isAlreadyRegistered()" check.
> 
> Well, I'd say that actually you could. :-)
> See that the reference to instance of SelendroidConfiguration is passed as a first parameter to SelfRegisteringRemote constructor.
> So just plug/unplug your devices and call performRegistration method to notify grid hub.
>
> (I actually use it for notifying grid hub if one of emulators crashed and that selendroid node has less free browsers while crashed emulator is rebuilding and restarting (this code is not published yet. not sure if it could be of use)).

I've got it. then, you will make another PR for this in the future, don't you?

selendroid-standalone/src/main/java/io/selendroid/standalone/server/grid/SelfRegisteringRemote.java
koichirok: I believe @frolix's case is more unusual...

@frolix wrote:
> I'd like to suggest to set the default value to 0 as it will be inline with @asm89 point above about disabling the recurrence behavior and with original grid's sources SelfRegisteringRemote.java, line 191.

this keeps behavior compatibility. :+1:

selendroid-standalone/src/main/java/io/selendroid/standalone/SelendroidConfiguration.java
asm89: Works for me. I'd add it to the documentation though (`0 to disable`). Should `0` also be the default value?

selendroid-standalone/src/main/java/io/selendroid/standalone/server/grid/SelfRegisteringRemote.java
asm89: @frolix selenium uses 30s?

selendroid-standalone/src/main/java/io/selendroid/standalone/server/grid/SelfRegisteringRemote.java
asm89: How does selenium handle this?

frolix:
@asm89, @koichirok, @DominikDary, please take a look. Tried to fix all the findings.

DominikDary:
:+1:

PR #764, opened 2014-12-30
Adding "deviceType" as a configurable parameter option.

binhvu:
So users can select a specific device together with the SDK version
for their tests. Previously, users can only specify the SDK version
for the tests and if there are several devices with the same SDK
version, selendroid will automatically select the first device to
carry out the test, which may not be the intended device for the
test.

This parameter can be ignored if the users are not desired to select
a specific device for the test.

Signed-off-by: Binh Vu <bvu@paypal.com>

### DIFF
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
@@ -63,6 +63,7 @@ public abstract class AbstractDevice implements AndroidDevice {
   private static final Logger log = Logger.getLogger(AbstractDevice.class.getName());
   public static final String WD_STATUS_ENDPOINT = "http://localhost:8080/wd/hub/status";
   protected String serial = null;
+  protected String model = null;
   protected Integer port = null;
   protected IDevice device;
   private ByteArrayOutputStream logoutput;
@@ -600,5 +601,9 @@ public abstract class AbstractDevice implements AndroidDevice {
   @Override
   public int hashCode() {
     return device.hashCode();
+  }
+
+  public String getModel() {
+    return model;
   }
 }
--- a/selendroid-standalone/src/test/java/io/selendroid/standalone/server/model/DeviceStoreFixture.java
+++ b/selendroid-standalone/src/test/java/io/selendroid/standalone/server/model/DeviceStoreFixture.java
@@ -17,10 +17,12 @@ import com.android.ddmlib.IDevice;
 
 import java.util.Map;
 
+import static io.selendroid.standalone.server.model.DeviceStoreFixture.withDefaultCapabilities;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 import io.selendroid.common.SelendroidCapabilities;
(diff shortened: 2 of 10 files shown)


selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
```diff
@@ -596,4 +597,8 @@ public boolean equals(Object o) {
   public int hashCode() {
     return device.hashCode();
   }
+
+  public String getDeviceType() {
+    return deviceType;
```
binhvu: Thanks @operatiay 
For real device, the deviceType is indicated by the hardware model. This hardware model is from the device itself and will be available after an initialization while adding the device to the device store.

operatiay: Thank you. I see that you're setting the deviceType to the value of the hardware model.
Is there any special reason to use a different naming?

binhvu: 1. Since <code>selendroid</code> already used <code>device</code> for for both emulator and real hardware as object under test.
2. <code>model</code> is only used in the limitation of real hardware object

Therefore I think <code>deviceType</code> is a logical choice to specify a different type of device to test if it is is an emulator, or a different type of model if it is a hardware device. Other than that, there is no special reasons to use <code>deviceType</code> in place of <code>model</code>.

operatiay: `device` for selendroid is either an emulator or real device.

based on 2. I'd prefer to use model for emulators as well and keep the terminology clear.
Implementation part is going to be reduced this way as well.

binhvu: @operatiay 
Agreed. I've modified the implementation to use <code>model</code> according to your recommendations. Thanks.

selendroid-standalone/src/test/java/io/selendroid/standalone/server/model/DeviceStoreFixture.java
```diff
@@ -32,6 +34,7 @@ protected static DefaultHardwareDevice anDevice(String name, DeviceTargetPlatfor
     DefaultHardwareDevice device = mock(DefaultHardwareDevice.class);
     when(device.getModel()).thenReturn(name);
     when(device.getTargetPlatform()).thenReturn(platform);
+    when(device.getModel()).thenReturn("Nexus 5");
```
asm89: The default value for this will be an empty string? If yes, can you change this to return an empty string instead and explicitly set the value to "Nexus 5" for the tests where you need that?

binhvu: @asm89 
I found that it is unnecessary to make that mock call (it is removed) since the original code already returned <code>name</code> with <code>getModel()</code>.

mach6:
:+1: thanks! this is merged

PR #768, opened 2015-01-03
fixes #767 stop internal ExecutorService of TrafficCounter on shutdown

### LINKED ISSUE #767: Threads keep running after SelendroidLauncher is stopped
Clean shutdown of jvm is not possible due to running threads after stopping SelendroidLauncher.

saral:


selendroid-server-common/src/main/java/io/selendroid/server/common/http/HttpServer.java
```diff
@@ -69,6 +69,7 @@ public void stop() {
       throw new IllegalStateException("Server is not running");
     }
     serverThread.interrupt();
+    TrafficCounter.shutdown();
```
asm89: Should this be called in `finally` above instead? The `TrafficCounter` is set up in the `ServerInitializer`.

saral: Yes it is probably better. I will refactor and commit. Thx.

asm89:
@saral Can you squash your commits?

This looks good to me. Can anyone else check as well? cc @selendroid/contributers

ainsej:
Yep, this fixes a bug I introduced, looks good to me!

Happy for it to be merged after squashing

PR #769, opened 2015-01-05
terminate must be called on AndroidDebugBridge on shutdown.

saral:
terminate must always be called on AndroidDebugBridge upon application termination for clean shutdown.

Note that terminate does not stop adb host server (only disconnectBridge does).

### DIFF
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/DefaultDeviceManager.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/DefaultDeviceManager.java
@@ -13,11 +13,10 @@
  */
 package io.selendroid.standalone.android.impl;
 
-import io.selendroid.standalone.android.AndroidDevice;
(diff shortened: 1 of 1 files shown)


asm89:
LGTM :+1:

smiklosovic:
@saral what is exactly "cleaned" by invoking terminate on AndroidDebugBridge?

saral:
In my trials, not calling terminate caused some background threads to remain running, preventing JVM to shutdown.
According to documentation of AndroidDebugBridge, terminate must be called upon application termination. I don't know what exactly is cleaned up, but I guess some internal Threads will be stopped, etc.

asm89:
You're right. See http://osxr.org/android/source/sdk/ddms/libs/ddmlib/src/com/android/ddmlib/AndroidDebugBridge.java#0202.

saral:
@smiklosovic Termination of application is when your application shuts down. I guess, you can reinit the Bridge with init(), if you need to. I don't know if double termination would cause problems. You need to look into documentation or try out.

PR #770, opened 2015-01-06
Removed Gson dependency from server and client

kpiwko:
* Replaced with JSONObject and JSONArray custom parsing

### DIFF
--- a/selendroid-server-common/src/main/java/io/selendroid/server/common/utils/CallLogEntry.java
+++ b/selendroid-server-common/src/main/java/io/selendroid/server/common/utils/CallLogEntry.java
@@ -56,12 +56,10 @@ public class CallLogEntry {
             entry.direction = json.optInt("direction", INCOMING_TYPE);
             entry.time = dateFormat.parse(json.optString("time"));
             return entry;
-        }
-        catch(ParseException e) {
+        } catch(ParseException e) {
             throw new IllegalArgumentException("Unable to parse CallLogEntry from string " + jsonString, e);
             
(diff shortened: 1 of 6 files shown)


lukeis:
it shouldn't be necessary to remove it from selendroid-client and it's nicer to use the more compact library there

selendroid-server-common/src/main/java/io/selendroid/server/common/utils/CallLogEntry.java
```diff
@@ -32,6 +48,38 @@ public CallLogEntry(String number, int duration, Date time, int direction) {
...
+    public String toJSON() {
+        try {
+            JSONObject json = new JSONObject();
+            json.put("number", number);
+            json.put("duration", duration);
+            json.put("direction", direction);
+            json.put("time", dateFormat.format(time));
+            return json.toString();
+        }
+        catch(JSONException e) {
```
asm89: Can you put the catch on the line above? `} catch(...) {` in other places as well.

asm89:
Can you fix the coding style? Thank you for fixing this. :+1:

DominikDary:
:+1:

kpiwko:
Sorry, github notifications were filter to wrong folder. Coding style was fixed.

kpiwko:
@lukeis I removed gson from the client as well to be sure that same code is used to marshal and demarshal JSON. For instance, with Date object it might be complicated to get it in GSON expected format from the custom utility.

PR #776, opened 2015-01-08
fix NPE on native apps driver.switchTo().defaultContent()

saral:
While testing a native app calling switchTo().defaultContent() on the Driver results in a NullPointerException.
This minor bug is fixed by this commit.

### DIFF
--- a/selendroid-server/src/main/java/io/selendroid/server/model/DefaultSelendroidDriver.java
+++ b/selendroid-server/src/main/java/io/selendroid/server/model/DefaultSelendroidDriver.java
@@ -13,49 +13,6 @@
  */
 package io.selendroid.server.model;
 
-import io.selendroid.server.ServerInstrumentation;
-import io.selendroid.server.android.AndroidTrackBall;
-import io.selendroid.server.android.AndroidWait;
-import io.selendroid.server.android.InstrumentedKeySender;
-import io.selendroid.server.android.KeySender;
-import io.selendroid.server.android.ViewHierarchyAnalyzer;
-import io.selendroid.server.android.WindowType;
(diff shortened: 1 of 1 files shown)


asm89:
@saral Wondering what the code path looks like that makes you end up in this state? The `DefaultSelendroidDriver` doesn't have a `selendroidWebdriver` at that point. Why does that happen?

I get that this PR "fixes" what you see, but I'm worried it isn't fixing the actual issue. If it is, should we log that there is no selendroid webdriver available?

saral:
`DefaultSelendroidDriver` has both a `selendroidWebDriver` and a `selendroidNativeDriver`. To my understanding one of these drivers is null, depending on your test. I tried to test a native app, so `selendroidWebDriver` was null.

In `DefaultSelendroidDriver` there are lots of methods which make a null check. But the null check was missing in `setFrameContext()`. 

Here is the code to reproduce (I have an Android device plugged in over usb):
```
SelendroidConfiguration config = new SelendroidConfiguration();
config.setShouldKeepAdbAlive(true);
config.addSupportedApp("/path/to/selendroid-test-app-0.13.0-SNAPSHOT.apk");

SelendroidLauncher launcher = new SelendroidLauncher(config);
launcher.launchSelendroid();

SelendroidCapabilities caps = 
  new SelendroidCapabilities("io.selendroid.testapp:0.13.0-SNAPSHOT");
WebDriver driver = new SelendroidDriver(caps);

driver.switchTo().defaultContent();

WebElement element = driver.findElement(By.xpath("//Button[@id='showToastButton']"));
element.click();
driver.quit();
```

DominikDary:
I'm fine mergin in the code, makes the tool more robust.

PR #777, opened 2015-01-08
Added support for Android 20 and 21

danderson1967:
Added support for Android 20 and 21 as well as tests to confirm that the max support version is Android 21.


asm89:
@danderson1967 Did you actually run tests on android 5.x?

danderson1967:
I ran my full suite of 140 tests on Android 5.0. However with the 0.13.0-SNAPSHOT codebase, I was unable to run the all of JUnit tests. Some were failing, and I didn't have the time to figure out why.

DominikDary:
I'm fine mergin this in, i have done it in the past like this. enabling it first and then adding the support. Then the users can at least use what is working fine.

PR #787, opened 2015-01-23
Adding "apiType" as an option in selendroid capabilities.

binhvu:
Since some applications are written for Google API, they would fail to
install on an emulator with Android API, and vice versa. This implementation
will add "apiType" option to the selendroid capabilties to allow users
to select a specific  emulator with the desired (Google/Android) API type to 
run their tests.

Users can safely ignore this option if the application under test is not
sensitive to either API type to run.

Signed-off-by: Binh Vu <bvu@paypal.com>

### DIFF
--- a/selendroid-common/src/main/java/io/selendroid/common/SelendroidCapabilities.java
+++ b/selendroid-common/src/main/java/io/selendroid/common/SelendroidCapabilities.java
@@ -48,6 +48,7 @@ public class SelendroidCapabilities extends DesiredCapabilities {
   public static final String PRE_SESSION_ADB_COMMANDS = "preSessionAdbCommands";
   public static final String SERIAL = "serial";
   public static final String MODEL = "model";
+  public static final String API_TYPE = "apiType";
 
   public static final String PLATFORM_VERSION = "platformVersion";
   public static final String PLATFORM_NAME = "platformName";
@@ -120,6 +121,10 @@ public class SelendroidCapabilities extends DesiredCapabilities {
     return (String) getRawCapabilities().get(SCREEN_SIZE);
   }
 
+  public String getAPIType() {
+    return (String) getRawCapabilities().get(API_TYPE);
+  }
+
   /**
    * Full path of the dex file (on the host machine) containing Selendroid extensions to be loaded at run time
    * Example: /home/user/extension.dex
@@ -142,6 +147,10 @@ public class SelendroidCapabilities extends DesiredCapabilities {
 
   public void setModel(String model) {
     setCapability(MODEL, model);
+  }
+
+  public void setAPIType(String apiType) {
+    setCapability(API_TYPE, apiType);
   }
 
   public void setPlatformVersion(DeviceTargetPlatform androidTarget) {
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/DefaultHardwareDevice.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/DefaultHardwareDevice.java
@@ -95,4 +95,8 @@ public class DefaultHardwareDevice extends AbstractDevice {
 	  return serial;
   }
 
+  @Override
+  public String getAPIType() {
+    return null;
+  }
 }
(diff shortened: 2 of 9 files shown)


selendroid-common/src/main/java/io/selendroid/common/SelendroidCapabilities.java
```diff
@@ -120,6 +121,10 @@ public String getScreenSize() {
     return (String) getRawCapabilities().get(SCREEN_SIZE);
   }
 
+  public String getAPIType() {
+      return (String) getRawCapabilities().get(API_TYPE);
```
mach6: trivial whitespace issue here.

selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/DefaultHardwareDevice.java
```diff
@@ -95,4 +95,8 @@ public String getSerial() {
 	  return serial;
   }
 
+  @Override
+  public String getAPIType() {
+    return null;
+   }
```
mach6: another trivial whitespace issue -- too much indent on the '}' character.

mach6:
I'm :+1: with this PR. Anyone else have feedback?

selendroid-common/src/main/java/io/selendroid/common/SelendroidCapabilities.java
binhvu: Fixed.

selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/DefaultHardwareDevice.java
binhvu: Fixed.

lukeis:
i'm not sure this is an appropriate noun to describe what this is?

Also this feels more like a boolean than a string value? Basically "isGoogleAPI", unless there's other variations?

lukeis:
(what I mean by the appropriate noun question, if you google "android api type" this is not what appears...)

mach6:
@lukeis I'm relying with @binhvu The trouble is when the device is a physical device neither `isGoogleAPI=true` or `isGoogleAPI=false` would apply since the information is not available when you query the device. Sure, you could do `isGoogleAPI=null`, but that seems sub-optimal too.

operatiay:
@mach6 if the information about installed GoogleAPI is needed on request this can be provided by the installed selendroid service (yes, you'd need to code in order to obtain the information). I agree with @lukeis  that the naming is not really appropriate for the purpose.

binhvu:
@mach6 @lukeis  @operatiay 
Since the queried information returned from the emulator categorized this as "target", the Android Virtual Device (AVD) Manager called this as "Target Name", can we named this as "targetName"?

operatiay:
what about something along `requiresGoogleAPI` with default being `false`?

binhvu:
"requiresGoogleAPI" sounds good to me. Boolean also good, since the hardware can return "false" instead null as default. @mach6  if you don't have any objection, I will revise the PR today.

binhvu:
This is about the name of the target API used in the emulator, it should be an identifier string, not a boolean. However, in the desired capabilities a boolean can be used to indicate the GoogleAPI is needed for the test or not. If "true" then "GoogleAPI" will be considered while selecting the emulator device by examine the above ID string, if "false" then any emulator devices can be a candidate. I still think "targetName" is good for the SelendroidCapabilities class to hold the name of the target API.

mach6:
@operatiay @lukeis @binhvu As we discussed, I'll take this PR from here. Furthermore, I'll pursue the design changes we talked about -- Mainly;

- [ ] Adding consistent a behavior when physical devices are attached to the node. This will likely mean an option passed to selendroid server which pre-validates each device for Google API support.
- [ ] Adding a new desired capability that tells selendriod to connect to a device which has the Google API (or other) available. This desired capability name will be more descriptive than `apiType` but will continue to be a `String` since we learned this is not simply a `true` (Google API present)/`false` (Google API not present) condition and there may support needed for additional API values in the near future.

Since the new work will originate from my fork, I'm closing this PR.

PR #799, opened 2015-02-03
Fix native view search with multiple top level views

asm89:
The current implementation of `AbstractNativeElementContext#searchViews`
can end up looking at multiple top level views. Selendroid makes sure
that the top level views passed to the `searchViews()` method are
ordered by when they were last painted on the screen. Searching for
views is then done by a breadth-first search to return elements higher
up in the view hierarchy first.

Given the two top level views below, the previous version of the
`searchViews()` method would return the following nodes in the order:
`[I, D, F, J, L]`.

```
        a                     g
     /     \              /     \
    b       c             h       I (1)
 /     \     \         /     \     \
D(2)    e     F (3)    J(4)   k     L (5)
```
*Capital letters are matching nodes.*

After this patch the method will return: `[D, F, I, J, L]`.
```
        a                     g
     /     \              /     \
    b       c             h       I (3)
 /     \     \         /     \     \
D(1)    e     F (2)    J(4)   k     L (5)
```

The behavior now changed from doing a breadth-first search spanning all
top level views at once, to doing it per view, returning the child
elements grouped per view.

This change has the biggest impact on finding one element. That will use
the same method, but return the first element of the list. In the
example above the value will change from `I' to `D`. This means you
always end up with a child element of the last drawn top level view,
instead of the child element that is highest up the view hierarchy
accross all views.

### DIFF
--- a/selendroid-server/src/main/java/io/selendroid/server/model/internal/AbstractNativeElementContext.java
+++ b/selendroid-server/src/main/java/io/selendroid/server/model/internal/AbstractNativeElementContext.java
@@ -268,28 +268,31 @@ public abstract class AbstractNativeElementContext
   }
 
 
-  public List<AndroidElement> searchViews(List<View> roots,
+  /** visible for testing */
+  protected static List<AndroidElement> searchViews(AbstractNativeElementContext context, List<View> roots,
       Predicate predicate, boolean findJustOne) {
     List<AndroidElement> elements = new ArrayList<AndroidElement>();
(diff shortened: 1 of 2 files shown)


asm89:
@DominikDary If you agree with this patch let's merge it. I filed a separate issue about why we're looking at multiple top level views at all. Let's discuss that case there: #800.

mkonicek:
:+1:

DominikDary:
:+1:

PR #801, opened 2015-02-03
Log click position

mkonicek:
This gives us more information about what is happening on the device when elements are clicked.


lukeis:
:+1:

DominikDary:
:+1:

PR #802, opened 2015-02-05
fix registration of selendroid server into a selenium grid

snordquist:
The SelendroidCapabilities requires the selendroid grid node to have the capability: platformName => android


DominikDary:
:+1:

PR #803, opened 2015-02-05
Reverted the 'adb.terminate()' call (back into if condition). 

operatiay:
As I've noticed in my testNG suite which uses Selendroid standalone, the commit f735a8a breaks the test execution of more than 1 test in a suite. With this change it works again.
The need for the original change was due to some orphaned processes upon test termination. This has to be addressed separately (and shouldn't break any already present functionality).

Also:
Simplified the logic of iterating through devices in DefaultDeviceManager#initializeAdbConnection()

### DIFF
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/DefaultDeviceManager.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/DefaultDeviceManager.java
@@ -84,11 +84,9 @@ public class DefaultDeviceManager extends Thread implements IDeviceChangeListene
           throw new RuntimeException(e);
         }
       }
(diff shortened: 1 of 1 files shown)


lukeis:
any way we can add a test to the test-app? (but I'm :+1: anyways)

DominikDary:
:+1:

operatiay:
A test could be added... but it's a testNG suite (this is where I was having troubles). Currently the tests are running using junit.
If none of you ( @DominikDary , @lukeis ) sees objections in adding another test runner, I'll add the test from here (https://gist.github.com/operatiay/6dc7c12dd2d76c95ef0c) and use a testng.xml file to execute it twice.

lukeis:
sgtm

PR #816, opened 2015-02-23
setAirplaneMode(true) for Android API 21

Blackbaud-KyleMartinez:
...nections activity. Using the default back button to get back to the app under test.

### DIFF
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
@@ -508,7 +508,7 @@ public abstract class AbstractDevice implements AndroidDevice {
   public void inputKeyevent(int value) {
     executeCommandQuietly(adbCommand("shell", "input", "keyevent", "" + value));
     // need to wait a beat for the UI to respond
-    sleep(600);
+    sleep(500);
   }
 
   public void invokeActivity(String activity) {
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
@@ -508,7 +508,7 @@ public abstract class AbstractDevice implements AndroidDevice {
   public void inputKeyevent(int value) {
     executeCommandQuietly(adbCommand("shell", "input", "keyevent", "" + value));
     // need to wait a beat for the UI to respond
-    sleep(500);
+    sleep(600);
   }
 
   public void invokeActivity(String activity) {
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/server/handler/NetworkConnectionHandler.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/server/handler/NetworkConnectionHandler.java
@@ -78,9 +78,14 @@ public class NetworkConnectionHandler extends BaseSelendroidStandaloneHandler {
 
 
     // nice and magical, this will 'close' the settings view and reopen the last application, the one under test! :)
-    device.inputKeyevent(KeyEvent.KEYCODE_DPAD_UP);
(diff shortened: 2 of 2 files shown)


Blackbaud-KyleMartinez:
When calling setAirplaneMode(true) while running an emulator using Android API 21 it will turn airplane mode on and then back off. This is because NetworkConnectionHandler is assuming that there will be a back button at the top of the Network Connection activity page. That button does not exist for this API level so I am using the devices main back button in order to get back to the app under test. I also increased the sleep time for running inputkeyevent() as I noticed 500 was not enough time. (I will be looking into a better way handling this. Using sleep is generally bad)

It would be nice to have a better way of getting back to the app under test. That would make this process a lot cleaner.

selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
```diff
@@ -508,7 +508,7 @@ public String runAdbCommand(String parameter) {
   public void inputKeyevent(int value) {
     executeCommandQuietly(adbCommand("shell", "input", "keyevent", "" + value));
     // need to wait a beat for the UI to respond
-    sleep(500);
+    sleep(600);
```
mmccartney: what does an extra 100 ms buy you?

selendroid-standalone/src/main/java/io/selendroid/standalone/server/handler/NetworkConnectionHandler.java
```diff
@@ -78,9 +78,14 @@ public Response handleRequest(HttpRequest request, JSONObject payload) throws JS
 
 
     // nice and magical, this will 'close' the settings view and reopen the last application, the one under test! :)
-    device.inputKeyevent(KeyEvent.KEYCODE_DPAD_UP);
-    device.inputKeyevent(KeyEvent.KEYCODE_DPAD_CENTER);
-
+    if (deviceAPILevel == 21) {
```
mmccartney: should this be `>= 21` ?

also, why are you sending *three* back keys?  Is there something special about 3?  should this be a loop looking for some specific behavior?  (assuming 3 back keys is right, please add a comment.)

selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
asm89: Please revert this, or explain why it improves things.

selendroid-standalone/src/main/java/io/selendroid/standalone/server/handler/NetworkConnectionHandler.java
Blackbaud-KyleMartinez: I updated the if statement to >= 21. 

3 back keys is the magical number here to get back to the app under test. There is no one action to take that will get you back to the app that I know of. The order of pages the process goes through is this.
App > Settings > network settings.

selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
Blackbaud-KyleMartinez: reverted

selendroid-standalone/src/main/java/io/selendroid/standalone/server/handler/NetworkConnectionHandler.java
lukeis: can you add a code comment to this effect about why we're hitting back 3 times

DominikDary:
@Blackbaud-KyleMartinez Can you please add to the code the comment why you use the three back button commands?
The reason why I ask is that for the future it is easier to understand.
If the comment is there, we will merge the code!

Blackbaud-KyleMartinez:
@DominikDary just added the code comments... Sorry for the delay.

DominikDary:
:+1: and thanks for your contribution!

PR #819, opened 2015-02-28
Pass Tests on Windows

chooper9:
A few tests would fail when building on a Windows machine due to the different newline character Windows uses and a file path in one test. Fixing these allows the project to build successfully on Windows without skipping tests.

### DIFF
--- a/selendroid-standalone/src/test/java/io/selendroid/standalone/android/impl/AbstractDeviceTest.java
+++ b/selendroid-standalone/src/test/java/io/selendroid/standalone/android/impl/AbstractDeviceTest.java
@@ -49,9 +49,15 @@ public class AbstractDeviceTest {
     when(device.getCrashLog()).thenCallRealMethod();
     when(
         device.executeCommandQuietly(argThat(matchesCmdLine("adb shell ls /storage/"))))  // The trailing '/' is key
-        .thenReturn("some_file\nappcrash.log\nanother_file");
+        .thenReturn(String.format("some_file%nappcrash.log%nanother_file"));
+    when(
+        device.executeCommandQuietly(argThat(matchesCmdLine("adb.exe shell ls C:\\storage/"))))  // The trailing '/' is key
+        .thenReturn(String.format("some_file%nappcrash.log%nanother_file"));
     when(
         device.executeCommandQuietly(argThat(matchesCmdLine("adb shell cat /storage/appcrash.log"))))
+        .thenReturn("crash log contents");
+    when(
+        device.executeCommandQuietly(argThat(matchesCmdLine("adb.exe shell cat C:\\storage\\appcrash.log"))))
         .thenReturn("crash log contents");
 
     assertEquals("crash log contents", device.getCrashLog());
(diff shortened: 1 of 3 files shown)


selendroid-standalone/src/test/java/io/selendroid/standalone/android/impl/AbstractDeviceTest.java
```diff
@@ -49,10 +49,16 @@ public void testGetCrashLogContents() {
     when(device.getCrashLog()).thenCallRealMethod();
     when(
         device.executeCommandQuietly(argThat(matchesCmdLine("adb shell ls /storage/"))))  // The trailing '/' is key
-        .thenReturn("some_file\nappcrash.log\nanother_file");
+        .thenReturn(String.format("some_file%nappcrash.log%nanother_file"));
+    when(
+        device.executeCommandQuietly(argThat(matchesCmdLine("adb.exe shell ls C:\\storage/"))))  // The trailing '/' is key
```
lukeis: file paths for adb shell commands should remain in unix style, since they are run on the android device and the file system is in that format

lukeis:
@mmccartney

chooper9:
In AbstractDevice.java, crashLogDirPath is set to crashLogFile.getParentFile().getAbsolutePath().
getAbsolutePath() determines the root with System.getProperty("user.dir"), which is returning C:\ on my system and causing the tests to fail. If running on an emulator that would return '/' properly, right? Should I just add
when(System.getProperty("user.dir")).thenReturn("/"); ?
As for the adb.exe, there's code in OS.java called from adbCommand() which explicitely returns that for windows: return isWindows() ? ".exe" : "";
What are your thoughts on fixing these?

selendroid-standalone/src/test/java/io/selendroid/standalone/android/impl/AbstractDeviceTest.java
mmccartney: are you saying this should be `adb.exe shell ls /storage/` instead?

could `matchesCmdLine()` take a regular expression and avoid the duplication of code?  for example:

```Java
    when(
         device.executeCommandQuietly(argThat(matchesCmdLine("adb(.exe)? shell ls /storage/"))))  // The trailing '/' is key
        .thenReturn(String.format("some_file%nappcrash.log%nanother_file"));
```

lukeis: and well, the theReturn should actually be with "/" and not %n substituted with path changes (since it should always return *nix style from an android device.

chooper9: Okay, I see. For some reason I'd assumed it used the system's (Windows, in my case) style. That just leaves the problem with
```
String crashLogDirPath = crashLogFile.getParentFile().getAbsolutePath();
```
because getAbsolutePath will call System.getProperty("user.dir") and return C:\ since the device is being mocked and the System is still Windows. It doesn't seem like you can mock java.lang.System. How would I get around that? Mocking the file class to change the result of getAbsolutePath? Accepting C:\ for the sake of the test? Neither of those seem like particularly nice solutions.

lukeis: well then... you've encountered another actual bug in selendroid's product! :-D  (yay for tests?)

AbstractDevice.java needs to be adjusted to always return unix style paths instead of unix and will need to be fixed.

PR #820, opened 2015-03-02
Add assets for inspector

mkonicek:
The `assets` directory contains all the JS and CSS files for the Selendroid inspector. The directory was missing in the APK since https://github.com/selendroid/selendroid/commit/7ae21137dde0b17522f08db20fb248ac0c3a9aa0.

This fixes the missing assets. 

The Selendroid inspector still crashes in JS in `onTreeLoaded`:

    this.root = this.jstree.jstree('get_json')[0];
    this.xml = this.root.metadata.xml;  // this.root is a div, metadata is undefined

I added at least some try-catch-finally error handling (this was missing) but I'm not familiar with the Inspector JS code. @DominikDary can you explain the purpose of `this.root.metadata.xml`? I don't see any recent changes to `inspector.js` or jQuery version bumps that could break this.

### DIFF
--- a/selendroid-server/assets/inspector/inspector.js
+++ b/selendroid-server/assets/inspector/inspector.js
@@ -146,25 +146,29 @@ Inspector.prototype.onNodeMouseOver = function (e, data) {
  * @param data
  */
 Inspector.prototype.onTreeLoaded = function (event, data) {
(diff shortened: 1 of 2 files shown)


lukeis:
:+1:

DominikDary:
thanks!

PR #823, opened 2015-03-03
Added apiTargetType as an option to SelendroidCapabilities to specify th...

ILikeToNguyen:
Added apiTargetType as an option to SelendroidCapabilities to specify that you need a device(emulator or physical) with Google APIs on it.

In addition to mock tests, I manually tested this works with emulators with google apis and without.  Also tested on physical device(samsung galaxy s2).


lukeis:
lgtm

mach6:
:+1: Thanks @lukeis and @ILikeToNguyen

PR #827, opened 2015-03-09
gets app version from pom.xml if class path doesn't start with jar

huntinglauren:
@lukeis 
updated (SelendroidServerBuilder.java) getJarVersionNumber() to parse pom.xml file to find version if classpath doesn't start with jar instead of just returning "dev"

### DIFF
--- a/selendroid-standalone/AndroidManifestTemplate.xml
+++ b/selendroid-standalone/AndroidManifestTemplate.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android" android:versionCode="1" android:versionName="0.13.0-SNAPSHOT" package="io.selendroid">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" android:versionCode="1" android:versionName="0.16.0-SNAPSHOT" package="io.selendroid">
 
     <uses-sdk android:minSdkVersion="10"/>
 


android-driver/project.properties
```diff
@@ -1,14 +0,0 @@
...
-#
-# To customize properties used by the Ant build system edit
-# "ant.properties", and override values to adapt the script to your
-# project structure.
-#
-# To enable ProGuard to shrink and obfuscate your code, uncomment this (available properties: sdk.dir, user.home):
-#proguard.config=${sdk.dir}/tools/proguard/proguard-android.txt:proguard-project.txt
-
-# Project target.
-target=android-16
```
lukeis: don't think you intended on removing this file?

selendroid-standalone/src/main/java/io/selendroid/standalone/builder/SelendroidServerBuilder.java
```diff
@@ -364,22 +360,77 @@ public static String getJarVersionNumber() {
...
-    } catch (Exception e) {
-      return "";
+      String dev_path = "";
+      if(classPath.startsWith("file") && classPath.contains("target"))
+        dev_path = classPath.substring(5, classPath.lastIndexOf("target"));
+      try {
+        version = getVersionFromPom(dev_path);
+      } catch (Exception e) {
+        e.printStackTrace();
+        return "pom file path may not exist";
```
lukeis: if this is the default fallback, probably should return "dev" instead.

selendroid-standalone/src/main/java/io/selendroid/standalone/builder/SelendroidServerBuilder.java
```diff
@@ -364,22 +360,77 @@ public static String getJarVersionNumber() {
...
 
+  private static String getVersionFromPom(String path) throws Exception {
+    String version = "";
+    path += "pom.xml";
+    String fileContents = getFileContents(path);
+
+    Pattern startPattern = Pattern.compile("<version>");
+    Matcher startMatch = startPattern.matcher(fileContents);
+
+    while(startMatch.find()) {
```
lukeis: let's use a simpler approach that doesn't include scanning the string, similar to this:

http://stackoverflow.com/a/336462  (this answer is doing more than we need here)

selendroid-standalone/src/main/java/io/selendroid/standalone/builder/SelendroidServerBuilder.java
```diff
@@ -364,22 +360,77 @@ public static String getJarVersionNumber() {
...
+        b.append(fileContents.charAt(i));
+        i++;
+      }
+      version = b.toString();
+      break;
+    }
+    return version;
+  }
+
+  private static String getFileContents(String file){
```
lukeis: we could probably use another util, like from CommonsIO and not have this method here.

http://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/FileUtils.html#readFileToString(java.io.File, java.nio.charset.Charset)

selendroid-standalone/src/main/java/io/selendroid/standalone/builder/SelendroidServerBuilder.java
huntinglauren: would something like this work? 
version = input.substring(input.indexOf(START_TAG) + START_TAG.length(), input.indexOf(END_TAG));

since we know the first <version> tag is the one we want, or should I grab the parent tag first and then find version within that just to be safe?

lukeis: i think using the regex pattern as shown in the link and getting the group (the value inbetween the tags, because it's surrounded by parenthesis) is a cleaner approach.

So, something like:
Pattern regex = Pattern.compile("<version>(.*?)</version>", Pattern.DOTALL);
Matcher matcher = regex.matcher(FileUtils.readFileToString(new File(path)));
if (matcher.find()) {
    return matcher.group(1);
}

mach6:
@huntinglauren I have two general comments;

1. Will you please squash your commits? A single commit makes following your PR a bit easier.
2. Perhaps I'm missing something but have you considered using a maven resource filter to populate the information (into a properties file, for example) you are after here?  The resource file would contain variables which maven will need to supply the values for at compile time -- when filtering the resources. Once the file is filtered, it will contain actual values and will become part of the jar. Next, it's simply a matter of loading the resource file via the classloader and exposing the values to the rest of the Java project as an Object. 

In this case, if you had a properties with an entry like `selendroid-standalone.build.version=selendroid-standalone-${project.version}`, the result after filtering would be `selendroid-standalone-0.16.0-SNAPSHOT` for a dev version or `selendroid-standalone-0.16.0` for a release version

lukeis:
@mach6 regarding (2) - she's trying to solve for when you are running tests from the IDE and it was just returning "dev" when executing selendroid's own tests.  The code already does look at the resource file in the jar, when it's run from the jar.

mach6:
@lukeis fair enough. :)

selendroid-standalone/src/main/java/io/selendroid/standalone/builder/SelendroidServerBuilder.java
mmccartney: how about parsing the XML using an XML parser instead?  This looks like a pretty easy to follow example:
http://www.mkyong.com/java/how-to-read-xml-file-in-java-dom-parser/

lukeis:
@huntinglauren closed? You can force push updates to your master fork and it will auto update this pull request

PR #829, opened 2015-03-11
User sends "true" for a Boolean DesiredCapability

ILikeToNguyen:
Be more forgiving with Boolean desiredCapabilities in the case where the user sends the boolean value over as a String like "true" instead of true.

### DIFF
--- a/selendroid-common/src/main/java/io/selendroid/common/SelendroidCapabilities.java
+++ b/selendroid-common/src/main/java/io/selendroid/common/SelendroidCapabilities.java
@@ -108,7 +108,7 @@ public class SelendroidCapabilities extends DesiredCapabilities {
   public Boolean getEmulator() {
     if (getRawCapabilities().get(EMULATOR) == null
         || getRawCapabilities().get(EMULATOR).equals(JSONObject.NULL)) return null;
-    return (Boolean) getRawCapabilities().get(EMULATOR);
+    return getBooleanCapability(EMULATOR);
   }
 
   public String getPlatformName() {
@@ -395,4 +395,17 @@ public class SelendroidCapabilities extends DesiredCapabilities {
     return listOfApps.size() > 0 ? listOfApps.last() : null;
   }
 
+  // throws exception if user didn't pass the capability as a boolean
+  private Boolean getBooleanCapability(String key) {
+    Object o = getRawCapabilities().get(key);
+    if (o == null) {
+      return null;
+    } else if (o instanceof Boolean) {
+      return (Boolean) o;
+    } else {
+      throw new ClassCastException(String.format(
+          "DesiredCapability %s's value should be boolean: found value %s of type %s", key, o.toString(), o.getClass()
+              .getName()));
+    }
+  }
 }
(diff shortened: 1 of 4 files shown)


lukeis:
not sure I agree with this change... users are used to (or should be used to) sending actual boolean values for boolean capabilities and we shouldn't be in the habit of allowing strings of booleans.

lukeis:
we could do better about any error messaging though, like responding with "this capability must be a boolean value"

ILikeToNguyen:
Sure, I can throw a better exception with a good error message instead of a classcast exception.

Sure you don't want to support bad practice? :) IOS-driver and sauce labs are forgiving and will read 'string true false' for boolean capabilities.

lukeis:
Yeah, I'd much prefer enforcing proper typing :)  Better error messages would be good.

selendroid-common/src/main/java/io/selendroid/common/SelendroidCapabilities.java
```diff
@@ -395,4 +395,16 @@ private String getDefaultVersion(Set<String> keys, String appName) {
...
 
+  // throws exception if user didn't pass the capability as a boolean
+  private Boolean getBooleanCapability(String key) {
+    Object o = getRawCapabilities().get(key);
+    if (o == null) {
+      return null;
+    } else if (o instanceof Boolean) {
+      return (Boolean) o;
+    } else {
+      throw new ClassCastException("DesiredCapability " + key
```
ILikeToNguyen: I can throw a different exception, i wanted to be more specific than just throwing a RuntimeException but couldn't really think of a better one. I thought about WebDriverException but not really sure if that fits.

The error message will look like this if user passes a string for boolean type of desiredcapability
DesiredCapability emulator's value is expected to be of type Boolean but has been set as type java.lang.String

mmccartney: this seems consistent with the exception that was previously thrown but has a more useful error message.  I think if you could add `o.toString()` in the message, that might be event more helpful:

`throw new ClassCastException(String.format("DesiredCapability %s's value should be boolean: found value %s of type %s", key, o.toString(), o.getClass().getName()))`

mach6:
@ILikeToNguyen Where does this PR stand?

ILikeToNguyen:
Pull request has been updated to reflect Mike's suggestion.  In the scenario when a user passes in a String value for a boolean desired capability, the exceptoin will look like this

Caused by: org.openqa.selenium.SessionNotCreatedException: DesiredCapability emulator's value should be boolean: found value true of type java.lang.String

selendroid-common/src/main/java/io/selendroid/common/SelendroidCapabilities.java
```diff
@@ -123,7 +123,7 @@ public String getLocale() {
     return (String) getRawCapabilities().get(LOCALE);
   }
 
-  public Map<String, Object> getRawCapabilities() {
+  private Map<String, Object> getRawCapabilities() {
```
mach6: @ILikeToNguyen Why the change in visibility here?

ILikeToNguyen: Just something small i noticed while i was in this class.  I couldnt find anything else in selendroid referencing this method so I thought maybe it was set as public on accident.  BUT you just made me realize, what if an end user is using this method instead of the getters for the specific selendroid capability they are querying.

Let me change it back to public since i dont want to break an external user that might be using this method already public.

mach6: @ILikeToNguyen sounds good. just ping me when this is updated.

PR #844, opened 2015-03-27
Clear data before app launch

JasnoWa:
Due to various conditions (race, user interaction between test runs, etc) the app isn't always in a clean state on launch with noClearData=false. Adding a data clear before the app is launched.

### DIFF
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/server/model/SelendroidStandaloneDriver.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/server/model/SelendroidStandaloneDriver.java
@@ -234,6 +234,10 @@ public class SelendroidStandaloneDriver implements ServerDetails {
           log.info("the app under test is already installed.");
         }
 
+        if(!serverConfiguration.isNoClearData()) {
+          device.clearUserData(app);
+        }
+
         int port = getNextSelendroidServerPort();
         boolean serverInstalled = device.isInstalled("io.selendroid." + app.getBasePackage());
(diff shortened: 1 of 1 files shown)


selendroid-standalone/src/main/java/io/selendroid/standalone/server/model/SelendroidStandaloneDriver.java
```diff
@@ -234,6 +234,10 @@ public String createNewTestSession(JSONObject caps, Integer retries) {
           log.info("the app under test is already installed.");
         }
 
+        if(!serverConfiguration.isNoClearData()) {
+          device.clearUserData(app);
```
asm89: This method call can throw an `AndroidSdkException`. Should we handle it here?

asm89:
@lukeis @JasnoWa Other issues/questions:
- can we also remove clearing the data after releasing the device? It always happens pre starting the app now anyway.
- can we update the docs here: https://github.com/selendroid/selendroid/blob/ec538326ccf930b9d981ec9cfdb555b795b4dc9a/selendroid-standalone/src/main/java/io/selendroid/standalone/SelendroidConfiguration.java#L106

asm89:
ping @JasnoWa @lukeis

asm89:
ping @JasnoWa @lukeis

JasnoWa:
Hey @asm89, Sorry for the super late delayed response (github email notifications were off). I left the clear data afterwards since I thought it should attempt to restore the app to the state it was at the start of the test.

I can create another pull request for the doc update if that's still desired.

PR #846, opened 2015-03-31
Drop explicit 'implements TakesScreenshot, JavascriptExecutor' declarations

sevaseva:
as RemoteWebDriver implements and will always implement those two.

Drop overriding of the getScreenshotAs() method as it is functionally
equivalent to the one in the parent class. (Note, this change is
independent from the one dropping 'implements' declarations).

This is a no-op, fully backwards compatible change and it will
simplify things for Selenium project in the future.


lukeis:
thanks @sevaseva can you sign the [cla](http://goo.gl/pAvxEI) for selendroid? (I don't believe google has signed one for the selendroid project)

sevaseva:
Signed

PR #848, opened 2015-04-04
Added touch gesture testing in selendroid-test app

colindmurray:
In collaboration with @chooper9, additional activites have been added to the selendroid-test app to support both automated and manual testing of touch gestures.

Detection of multi-touch is also included, which hopefully will be helpful in testing future extensions to selendroid's multi-touch support.


lukeis:
I don't mind you guys having a few commits (to show who's done what). But can you squash your commits and rebase? This will be a little bit tricky here. But I would like to see at most 7 commits here instead of the 25. Of course you could also re-order the commits and just have 2 commits (if that will even work... don't sweat it if it doesn't).

colindmurray:
Will, do.

colindmurray:
I've squashed it down to really the important 6 commits.

DominikDary:
:+1:

PR #852, opened 2015-04-15
Hardware devices apiTargetType need to default to android

ILikeToNguyen:


mach6:
:+1: LGTM

asm89:
@ILikeToNguyen @mach6 Why not set the property in the base class to the default value? https://github.com/selendroid/selendroid/blob/932eba6956c52ccf5a5a7cea795f7e09227dead5/selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java#L66

ILikeToNguyen:
@asm89 You're right I could have also set the default value at variable declaration in the base class.

asm89:
@ILikeToNguyen Can you send a PR to fix that?

PR #860, opened 2015-04-24
Action chain

h2dennis:
Worked in collaboration with @chooper9 @colindmurray 
Refactored SafeHandle for future multitouch implementation.
Previous implementation didn't properly handle thread sleep for synchronizing touch events.

### DIFF
--- a/selendroid-server/src/main/java/io/selendroid/server/handler/Actions.java
+++ b/selendroid-server/src/main/java/io/selendroid/server/handler/Actions.java
@@ -5,6 +5,7 @@ import java.util.List;
 
 import io.selendroid.server.action.ActionContext;
 import io.selendroid.server.action.ActionHandler;
+import io.selendroid.server.action.ActionChain;
 import io.selendroid.server.common.Response;
 import io.selendroid.server.common.SelendroidResponse;
 import io.selendroid.server.common.action.touch.TouchActionName;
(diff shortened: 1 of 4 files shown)


h2dennis:
@DominikDary  @lukeis 
Guys, is there anything wrong with the code?

DominikDary:
I just did not had the time to review.

selendroid-server/src/main/java/io/selendroid/server/handler/Actions.java
```diff
@@ -64,51 +43,49 @@ public Response safeHandle(HttpRequest request) throws JSONException {
...
+            handler.handle(actionName, getSelendroidDriver(request), action, chain.getContext());
           }
         }
       }
 
-      int pauseTime = longestPause(i, payload);
-      if (pauseTime > 0) {
+      // Sleep the thread until the next non-pause action
+      if(allPaused) {
+        int pauseTime = shortestPause(actionChains);
```
lukeis: so, this shouldn't be the shortest pause, we should keep the original 'longest' pause. The intention of multi-action chains is to gate each 'tick' with all the other actions at it's tick.

    [
      [down, pause 10, release],
      [pause 0, pause 1, tap]
    ]

This should not result in the tap being performed 9 seconds before the release. This example payload should perform the "down" command, then a 10 second pause (the longest) in the 'tick', then release and tap at the "same" time (the release would get called first, then the tap)

Also... both original and as proposed are flawed with finding what the sleep interval should be. It should be determined in the current 'tick' of the action chain, not searching the entire payload (or actionChains) and using that time.

colindmurray:
Are there more instances other than multitouch that use multi-action chains?  I had assumed mutli-action chains was attempting to address multi-touch.  In the below code for multitouch I assume the intent is the actions execute in parallel, regardless of differences in pause times?
```
TouchAction finger1 = new TouchActionBuilder().pointerDown().pause(500).
  pointerMove(x, y).pointerUp().build(); 
TouchAction finger2 = new   TouchActionBuilder().pointerDown().pause(100).
  pointerMove(x, y).pointerUp().build(); 
MultiTouchAction multiAction = new MultiTouchAction(finger1, finger2); 
multiAction.perform(driver);```

In this case, I'd expect as a user finger2 to finish first, then finger1.  In order to achieve this we would need to sleep for the shortest pause (100ms), then send a MotionEvent that describes which pointers changed and which did not before continuing the pause (400 remaining ms).  If what I understand is correct for multitouch, we could work around longest pause.  Instead we could try identifying if the request being handled is a multitouch action prior to iterating on the action chains.

lukeis:
No... in order to accomplish something like you're suggesting then the chains should look more like this:

    t1 = TouchAction.pointerDown.pause(100).pointerMove(x,y).pause(400).pointerUp();
    t2 = TouchAction.pointerDown.pause(100).pointerMove(x,y).pointerUp();
    MultiTouchAction(t1,t2).perform();

We should be preserving 'tick' order and executing each 'tick' and the Selenium project will need to document this better? Or we can also try to document this behavior too. This is defined in the spec:
https://w3c.github.io/webdriver/webdriver-spec.html#parallel-actions-1

chooper9:
Oh, I see. When we saw 'these actions will be executed in parallel' on the Selendroid Mobile Gestures page, we thought it meant the execution of them would be as if done on separate threads - completely independent of one another. It might help to instead say something like, 'these will be executed as parallel actions, as defined by the WebDriver API' and maybe even include that link or put a section for Parallel Actions in the Advanced User Interactions API.

chooper9:
Changed to function as per the spec.

lukeis:
lgtm now!

@h2dennis can you sign the [cla](http://goo.gl/pAvxEI). I see the others on this PR already have :)

h2dennis:
@lukeis The cla is signed now.

Issue #752, opened 2014-12-15
[Enquiry] Unlock screen before test


Hassan-Radi:
I am using Selendroid 0.10. I noticed that Selendroid doesn't unlock the device screen automatically by itself. I was wondering if this was a bug or a non-implemented feature??

asm89:
I'm closing this issue due to inactivity. If you still experience the issue described on the latest release of selendroid feel free to file a new issue. If you do please provide a [concise reproducible test case](http://sscce.org/) and describe what results you are seeing and what results you expect.

PR #861, opened 2015-04-27
Added code to unlock phone before running tests

mtimkovich:
#752


lukeis:
I assume this is meant to be for a 'real device' rather than emulator?

The emulator code is done in [startAndroidEmulator](https://github.com/selendroid/selendroid/blob/master/selendroid-standalone/src/main/java/io/selendroid/standalone/server/model/SelendroidStandaloneDriver.java#L370). So maybe we should do this [above where that is called](https://github.com/selendroid/selendroid/blob/master/selendroid-standalone/src/main/java/io/selendroid/standalone/server/model/SelendroidStandaloneDriver.java#L226-L228) in an else clause?

PR #862, opened 2015-04-29
Delete temporary created by Selendroid Server

nadaismail:
Added a command line flag, -deleteTmpFiles, to delete temporary files that persist in the temp directory. Files were being created after Selendroid Standalone Server was started and test app was run. When server is started with flag, then files will be deleted when server is stopped.

### DIFF
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/SelendroidConfiguration.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/SelendroidConfiguration.java
@@ -140,6 +140,9 @@ public class SelendroidConfiguration {
   @Parameter(names = "-folder", description = "The folder which contains Android applications under test. This folder will monitor and add new apps to the apps store during the lifetime of the selendroid node.")
   private String folder = null;
 
+  @Parameter(names = "-deleteTmpFiles", description = "Deletes temporary files created by the Selendroid Server.")
+  private boolean deleteTmpFiles = true;
+
   public void setKeystore(String keystore) {
     this.keystore = keystore;
   }
@@ -362,6 +365,14 @@ public class SelendroidConfiguration {
 
   public String getAppFolderToMonitor() {
     return folder;
+  }
+
+  public boolean isDeleteTmpFiles() {
+    return deleteTmpFiles;
+  }
+
+  public void setDeleteTmpFiles(boolean deleteTmpFiles){
+    this.deleteTmpFiles = deleteTmpFiles;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


selendroid-standalone/src/main/java/io/selendroid/standalone/server/model/SelendroidStandaloneDriver.java
```diff
@@ -164,8 +164,12 @@ public void addToAppsStore(File file) throws AndroidSdkException {
         // using "android" as the app name, because that is the desired capability default in
         // selenium for
         // DesiredCapabilities.ANDROID
+          File androidAPK = androidDriverAPKBuilder.extractAndroidDriverAPK();
+          if(serverConfiguration != null && serverConfiguration.isDeleteTmpFiles()) {
+              androidAPK.deleteOnExit(); //Deletes temporary files if flag set
```
lukeis: seems like something we should just have on by default? since the deleteOnExit seems pretty safe, since it'll wait for JVM to exit.

and we probably don't need a configuration flag around it then

lukeis:
also, can you please update your edits to the style format used in selendroid (2 spaces for tabs). And can you rebase your change? (as to not have the merge commits included?)

selendroid-standalone/src/main/java/io/selendroid/standalone/server/model/SelendroidStandaloneDriver.java
lukeis: actualy, i take that back, the config should stay... but we should default it to true and let people disable this as the exception to the rule :)

selendroid-standalone/src/main/java/io/selendroid/standalone/SelendroidConfiguration.java
```diff
@@ -140,6 +140,9 @@
   @Parameter(names = "-folder", description = "The folder which contains Android applications under test. This folder will monitor and add new apps to the apps store during the lifetime of the selendroid node.")
   private String folder = null;
 
+    @Parameter(names = "-deleteTmpFiles", description = "Deletes temporary files created by the Selendroid Server.")
+    private boolean deleteTmpFiles = false;
```
lukeis: as per my other comments, can you change this to true :)

selendroid-standalone/src/main/java/io/selendroid/standalone/server/model/SelendroidStandaloneDriver.java
DominikDary: :+1:

nadaismail:
I updated my edits to the style format used in Selendroid and set -deleteTmpFiles to true by default

lukeis:
whoops, @nadaismail can you please sign the cla?

http://goo.gl/pAvxEI

PR #867, opened 2015-05-05
Wake screen before running tests

Mdelvalle:
On Android devices, if the screen is off, tests would not run; this turns the device on before running tests.  
#752   
@mtimkovich and I worked on this.  
Should I squash the commits?

### DIFF
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/android/AndroidDevice.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/android/AndroidDevice.java
@@ -96,4 +96,6 @@ public interface AndroidDevice {
   public String getModel();
 
   public String getAPITargetType();
+
+  public void unlockScreen() throws AndroidDeviceException;
 }
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/android/AndroidEmulator.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/android/AndroidEmulator.java
@@ -54,7 +54,7 @@ public interface AndroidEmulator {
 
   public void setSerial(int port);
 
-  public void unlockEmulatorScreen() throws AndroidDeviceException;
+  public void unlockScreen() throws AndroidDeviceException;
 
   public void setWasStartedBySelendroid(boolean wasStartedBySelendroid);
 
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/DefaultAndroidEmulator.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/DefaultAndroidEmulator.java
@@ -354,7 +354,7 @@ public class DefaultAndroidEmulator extends AbstractDevice implements AndroidEmu
 
     log.info("Emulator start took: " + (System.currentTimeMillis() - start) / 1000 + " seconds");
     log.info("Please have in mind, starting an emulator takes usually about 45 seconds.");
-    unlockEmulatorScreen();
+    unlockScreen();
 
     waitForLauncherToComplete();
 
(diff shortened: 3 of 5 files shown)


selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
```diff
@@ -607,4 +607,31 @@ public String getModel() {
   public String getAPITargetType() {
     return apiTargetType;
   }
+
+  public void unlockScreen() throws AndroidDeviceException {
+    // Get phone's android version and whether the screen is off.
```
lukeis: I agree this method should be defined in AbstractDevice, but the contents you have here should go into DefaultHardwareDevice

and then the method unlockEmulatorScreen in DefaultAndroidEmulator should get renamed.

selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
```diff
@@ -607,4 +607,31 @@ public String getModel() {
...
+    // Different ways to detect if the screen is off depending on the version.
+    CommandLine commandVersion = adbCommand("shell", "getprop", "ro.build.version.release");
+    String version = executeCommandQuietly(commandVersion);
+    CommandLine commandScreen = adbCommand("shell", "dumpsys", "power");
+    String screen = executeCommandQuietly(commandScreen);
+
+    // Kitkat
+    if (version.matches("^4\\.4(\\..*)?")) {
```
lukeis: i see you've only handled KitKat and Lollipop (likely cause that's all you have access to and I don't have access to older either right now)... but selendroid attempts to support back to Gingerbread (2.3.3 / API level 10). Do we have any idea what it may be there and the versions inbetween? If we can't tell, let's just try to have it do the same thing the KitKat one does here and maybe others with devices on the old versions can either verify or fix it later :)

lukeis:
also, there's no need to close a pull request in case you're attempting to rebase a commit. You can make your changes and force push changes to your github branch (then this pull request would get automatically updated)

selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
```diff
@@ -607,4 +607,31 @@ public String getModel() {
   public String getAPITargetType() {
     return apiTargetType;
   }
+
+  public void unlockScreen() throws AndroidDeviceException {
+    // Get phone's android version and whether the screen is off.
+    // Different ways to detect if the screen is off depending on the version.
+    CommandLine commandVersion = adbCommand("shell", "getprop", "ro.build.version.release");
```
lukeis: can you use this.getTargetPlatform() instead? And then just check targetPlatform.getVersionNumber() or .getApi() (parseInt and check for > 19 for lollipop)

selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
Mdelvalle: Okay, I'll do that.
What should unlockEmulatorScreen get renamed to?

selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
Mdelvalle: No idea what the difference might be for versions before KitKat, but I'll try to find out.
If I can't figure it out, I'll do what you said.

Mdelvalle:
Ah, you caught me!

Most of those commits are just pulls from the main remote branch.
I just wasn't sure if there would be any conflicts because of that.

lukeis:
should be fine, conflicts would have shown up on the merge commits too. just do:

    git pull --rebase <selendroid-named-remote> master

and replace "<selendroid-named-remote>" with what's appropriate for you :)

selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
lukeis: unlockScreen  :)

lukeis:
looks like you've taken care of most of this, can you please rebase your commit (git pull --rebase) and can you update AbstractDevice.java (removing the contents of unlockScreen)

lukeis:
awesome, thanks for cleaning that up! :+1:

Mdelvalle:
No problem, let me know if there's anything else I can do!

PR #868, opened 2015-05-06
Multitouch

chooper9:
Worked in collaboration with @colindmurray.
Implemented multitouch functionality usable from creating a MultitouchAction with multiple TouchActions.

### DIFF
--- a/selendroid-server/src/main/java/io/selendroid/server/action/ActionChain.java
+++ b/selendroid-server/src/main/java/io/selendroid/server/action/ActionChain.java
@@ -18,10 +18,11 @@ public class ActionChain {
   private int pauseTime = 0;
   private int index = 0;
 
-  public ActionChain(JSONObject actionChain) throws JSONException {
+  public ActionChain(JSONObject actionChain, int i) throws JSONException {
     this.inputDevice = actionChain.getString("inputDevice");
     this.actionChain = actionChain.getJSONArray("actions");
     this.context = new ActionContext();
+    this.context.setId(i);
   }
 
   public int length() {
--- a/selendroid-server/src/main/java/io/selendroid/server/action/ActionChain.java
+++ b/selendroid-server/src/main/java/io/selendroid/server/action/ActionChain.java
@@ -18,11 +18,13 @@ public class ActionChain {
   private int pauseTime = 0;
   private int index = 0;
 
-  public ActionChain(JSONObject actionChain, int i) throws JSONException {
+  private static int id_counter = 0;
+
+  public ActionChain(JSONObject actionChain) throws JSONException {
     this.inputDevice = actionChain.getString("inputDevice");
     this.actionChain = actionChain.getJSONArray("actions");
     this.context = new ActionContext();
-    this.context.setId(i);
+    this.context.setId(id_counter++);
   }
 
   public int length() {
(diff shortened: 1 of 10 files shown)


selendroid-server/src/main/java/io/selendroid/server/action/ActionChain.java
```diff
@@ -18,10 +18,11 @@
   private int pauseTime = 0;
   private int index = 0;
 
-  public ActionChain(JSONObject actionChain) throws JSONException {
+  public ActionChain(JSONObject actionChain, int i) throws JSONException {
     this.inputDevice = actionChain.getString("inputDevice");
     this.actionChain = actionChain.getJSONArray("actions");
     this.context = new ActionContext();
+    this.context.setId(i);
```
chooper9: Each ActionChain needs a unique identifier to correspond with a pointer in the MotionEvent. When we implemented this we just set the id when creating the actionchain from Actions.java but only recently realized that the JSONObject representing an actionchain has a placeholder for 'id' which is set in MultitouchScreen in selendroid client. Would it be better to use that? It's currently unused and I believe this is its intended purpose but wasn't completely sure.

lukeis: seems like these id's could get accidentally re-used with subsequent calls and could have some potential adverse affects?

Why don't we use a UUID and set it here?
http://developer.android.com/reference/java/util/UUID.html#randomUUID()

Then the method signature can go back to just taking a JSONObject

chooper9: Sounds good. And just to be clear: we should generate a UUID here in the ActionChain constructor - not in MultitouchScreen?

lukeis: no, i don't think in multitouchscreen is a good idea, the "id" placeholder in there appears to be for the entire multi touch action chain(s).

It seems as though this ID needs to be unique per chain event, to retain the correct pointer event to fire combinations with.

chooper9: As a disclaimer, I haven't used UUID's before so I could be wrong about this. From what I've read, UUID's are 128 bits but the unique pointer id used by Android's MotionEvents is an int. I don't think UUID's can be used to provide a unique integer. As an alternative, a class variable might work. What are your thoughts?

colindmurray: We could use [UUID's hashcode](http://www.tutorialspoint.com/java/util/uuid_hashcode.htm).  For the purposes here, the 32 bits of randomness should be enough.  The one in many million chance that a collision does happen it might make for a confusing test run.

Incrementing a class variable initialized to 0 perhaps could work as @chooper9 mentioned also.

selendroid-server/src/main/java/io/selendroid/server/action/ActionChain.java
```diff
@@ -18,10 +18,13 @@
   private int pauseTime = 0;
   private int index = 0;
 
+  private static int id_counter = 0;
```
chooper9: Updated ActionChain to assign id via a class variable to avoid reusing ids.

lukeis:
lgtm, thanks!

Issue #876, opened 2015-05-20
Selendroid keep waiting for emulator to be closed eternally


varyvol:
When test session finishes and Selendroid tries to close the emulator, it is actually closed but lock files in .android/$avd/ are not removed. Therefore, Selendroid does not detect the emulator is closed and keep waiting eternally.

PR #877, opened 2015-05-21
When emulator killed, remove lock files to avoid infinite loop

varyvol:
Solves Issue #876 

When emulator cannot be reached through Telnet, then it was closed and therefore lock files are removed.

### DIFF
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/DefaultAndroidEmulator.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/DefaultAndroidEmulator.java
@@ -464,15 +464,21 @@ public class DefaultAndroidEmulator extends AbstractDevice implements AndroidEmu
   }
 
   private void stopEmulator() throws AndroidDeviceException {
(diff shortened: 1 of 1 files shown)


selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/DefaultAndroidEmulator.java
```diff
@@ -494,6 +500,8 @@ public void stop() throws AndroidDeviceException {
           try {
             stopEmulator();
           } catch (AndroidDeviceException sce) {
+	    removeLockFiles();
```
lukeis: can you fix the spacing here, also don't need a blank line afterwards.

lukeis: also there's a 'stopEmluator' call earlier in the method that could potentially throw the AndroidDeviceException too now. You'll want to catch it there too.

asm89:
ping @varyvol

PR #883, opened 2015-06-03
Added support for Android 22 and modified max value test

migchez:
Test Plan: Ran tests on Android 22. Tests failed. Updated jars experimentally. Ran tests. Tests passed.


selendroid-common/src/test/java/io/selendroid/common/device/DeviceTargetPlatformTest.java
```diff
@@ -17,13 +17,9 @@ public void shouldLoadFromStringAndNumberValue() {
     Assert.assertEquals(DeviceTargetPlatform.ANDROID17, platform);
   }
   @Test
-  public void shouldLoadFromMaxSupportedNumberValue() {
-    DeviceTargetPlatform platform = DeviceTargetPlatform.fromPlatformVersion("21");
-    Assert.assertEquals(DeviceTargetPlatform.ANDROID21, platform);
+  public void shouldLoadMaxValue() {
```
asm89: Remove this test. It doesn't make sense.

PR #888, opened 2015-06-11
Move DecorViewPredicate to own class and add to PredicateFactory

migchez:
Simply moved DecorViewPredicate into its own file and adds it to the PredicateFactory.

Test plan:
mvn clean compile package


selendroid-server/src/main/java/io/selendroid/server/model/TagNamePredicate.java
```diff
@@ -28,3 +28,4 @@ public boolean apply(View to) {
     return to.getClass().getSimpleName().equals(using);
   }
 }
+
```
asm89: nit: can you revert the change in this file?

asm89:
:+1:

Issue #892, opened 2015-06-13
Please allow a true/Boolean capability to be defined with the string "true"


sglebs:
I am trying to use Selendroid from FitNesse https://github.com/fhoeben/hsac-fitnesse-fixtures/issues/10 but it passes "true" down. Can we please change this code below (in SelendroidCapabilities) so that it also takes the string "true" as a valid boolean/true value?

```
  public Boolean getEmulator() {
    if (getRawCapabilities().get(EMULATOR) == null
        || getRawCapabilities().get(EMULATOR).equals(JSONObject.NULL)) return null;
    return getBooleanCapability(EMULATOR);
  }

  // throws exception if user didn't pass the capability as a boolean
  private Boolean getBooleanCapability(String key) {
    Object o = getRawCapabilities().get(key);
    if (o == null) {
      return null;
    } else if (o instanceof Boolean) {
      return (Boolean) o;
    } else {
      throw new ClassCastException(String.format(
          "DesiredCapability %s's value should be boolean: found value %s of type %s", key, o.toString(), o.getClass()
              .getName()));
    }
  }
```

Thanks for listening!

lukeis:
There's other Boolean capabilities and other drivers also have Boolean capabilities. I'd highly recommend updating your library to allow sending Boolean capabilities as there's no guarantee other new ones or other drivers won't require them in the future.

fhoeben:
Why don't you use Capabilities `is` method instead of having your own `getBooleanCapability`? Or at least copy its behavior?

http://javadox.com/org.seleniumhq.selenium/selenium-api/2.40.0/org/openqa/selenium/Capabilities.html#is(java.lang.String)

```java
  public boolean is(String capabilityName) {
    Object cap = getCapability(capabilityName);
    if (cap == null) {
      return false;
    }
    return cap instanceof Boolean ? (Boolean) cap : Boolean.parseBoolean(String.valueOf(cap));
  }

```

sglebs:
I like @fhoeben 's suggestion. Less code and more versatile. Great!

sglebs:
Great, @fhoeben ! I liked the "This pull request can be automatically merged by project collaborators.". I hope it gets incorporated soon! :)

lukeis:
merged the PR

PR #894, opened 2015-06-14
Allow string value for boolean capabilities

fhoeben:
Address #892: allow "false" or "true" as String value for boolean capabilities…

I have signed the CLA.

### DIFF
--- a/selendroid-common/src/main/java/io/selendroid/common/SelendroidCapabilities.java
+++ b/selendroid-common/src/main/java/io/selendroid/common/SelendroidCapabilities.java
@@ -395,17 +395,13 @@ public class SelendroidCapabilities extends DesiredCapabilities {
     return listOfApps.size() > 0 ? listOfApps.last() : null;
   }
 
-  // throws exception if user didn't pass the capability as a boolean
+  // throws exception if user didn't pass the capability as a boolean or String parsable as boolean
   private Boolean getBooleanCapability(String key) {
     Object o = getRawCapabilities().get(key);
     if (o == null) {
       return null;
-    } else if (o instanceof Boolean) {
-      return (Boolean) o;
     } else {
-      throw new ClassCastException(String.format(
-          "DesiredCapability %s's value should be boolean: found value %s of type %s", key, o.toString(), o.getClass()
-              .getName()));
+      return is(key);
     }
   }
 }
--- a/selendroid-common/src/main/java/io/selendroid/common/SelendroidCapabilities.java
+++ b/selendroid-common/src/main/java/io/selendroid/common/SelendroidCapabilities.java
@@ -400,8 +400,16 @@ public class SelendroidCapabilities extends DesiredCapabilities {
     Object o = getRawCapabilities().get(key);
     if (o == null) {
       return null;
+    } else if (o instanceof Boolean) {
+      return (Boolean) o;
+    } else if (o instanceof String
+            && ("true".equalsIgnoreCase((String) o)
+            || "false".equalsIgnoreCase((String) o))) {
+      return Boolean.valueOf((String) o);
     } else {
-      return is(key);
+      throw new ClassCastException(String.format(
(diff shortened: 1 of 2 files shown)


selendroid-common/src/main/java/io/selendroid/common/SelendroidCapabilities.java
```diff
@@ -395,17 +395,13 @@ private String getDefaultVersion(Set<String> keys, String appName) {
...
     Object o = getRawCapabilities().get(key);
     if (o == null) {
       return null;
-    } else if (o instanceof Boolean) {
-      return (Boolean) o;
     } else {
-      throw new ClassCastException(String.format(
-          "DesiredCapability %s's value should be boolean: found value %s of type %s", key, o.toString(), o.getClass()
-              .getName()));
+      return is(key);
```
lukeis: I'm in favor of this change, but can we catch the exception that can occur when the value isn't boolean and passed into "Boolean.parseBoolean(String.valueOf(cap))"

Adding a test case of cap.put("emulator", "kitkat");  or something of the like and seeing the error message bubble up accordingly

fhoeben: No exception would be thrown, the value would be 'false'. Is that acceptable/desirable?

From `Boolean.class`
```java
    private static boolean toBoolean(String name) {
        return ((name != null) && name.equalsIgnoreCase("true"));
    }
```

lukeis: hmm, that method makes somewhat sense in core java, but I'm not sure I like that behavior for selendroid. We're expecting a boolean value, but essentially we only care about either true or "true" coming across... everything else is false. Is it going to be confusing to any users?

I'd almost rather add an additional else clause to check for the string of "true" or "false", otherwise raise the exception.

Since users of other programming languages (like javascript or python) may have a different impression of what a truthy value should be.

fhoeben: I understand your point, but by using the core Selenium behavior there will be consistent behavior across web drivers. I believe that will be quite good.

lukeis: problem is, i'm still here to listen to user complaints / issues later on. And I much prefer error messages with explicit strings attached than something that essentially is a side-effect.

So, I'll merge it if you change to checking explicitly for 'true' or 'false' strings in addition to the previous Boolean instance check, but I don't want the is() behavior that selenium core library provides as it doesn't correct users doing things poorly (and if we did actually use the is(), we'd need some documentation somewhere to explain what the side effects are... otherwise, more confused users, more noisy issues).

Add on top to the fact that this change is only getting in place because some selenium library doesn't allow you to pass actual boolean values as capabilities (which is a bug in that library :) ).

fhoeben: I get your point

lukeis:
thanks @fhoeben :)

selendroid-common/src/test/java/io/selendroid/common/SelendroidCapabilitiesTest.java
```diff
@@ -42,6 +45,61 @@ public void testInstantiateFromJSON() throws Exception {
...
+      SelendroidCapabilities capa = new SelendroidCapabilities(map);
+      boolean em = capa.getEmulator();
+      Assert.fail("Expected exception, got: " + em);
+    } catch (ClassCastException e) {
+      String msg = e.getMessage();
+      Assert.assertTrue("Expected key in message: " + msg, msg.contains("emulator"));
+      Assert.assertTrue("Expected value in message: " + msg, msg.contains("kitkat"));
+      Assert.assertTrue("Expected class in message: " + msg, msg.contains("String"));
+    }
+  }
```
asm89: :+1: yay for tests

PR #904, opened 2015-06-23
Various fixes

frolovs:
- Add WXGA720 skin for detection of screen resolution.
- Optimize screenshots from standalone server - use 3 bytes per pixel instead of 4 bytes. Produces 25% smaller BufferedImage object.
- Fix bug SelendroidServerBuilder when executing tests and using default keystore alias and passwords and androidKeyStore file exists - initialize fields with default values.


lukeis:
:+1:  looks ok to me

asm89:
:+1:

Issue #855, opened 2015-04-20
selendroidDriver.readCallLog() fails with ClassCastException


NamrataBagerwal:
Hi all, while getting call log details I am getting the following exception. 

java.lang.ClassCastException: java.util.ArrayList cannot be cast to java.lang.String
	at io.selendroid.client.SelendroidDriver.readCallLog(SelendroidDriver.java:255)
	at com.spec.selendroidtestappdemo.nativetests.AddCallLogTest.addCallLogTest(AddCallLogTest.java:60)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.lang.reflect.Method.invoke(Unknown Source)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:271)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:70)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:50)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:238)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:63)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:236)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:53)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:229)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:309)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:50)
	at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:38)
[...]

PR #916, opened 2015-06-29
[client] Fix SelendroidDriver#readCallLog()

asm89:
AddCallLogTest was broken. Turns out the `readCallLog()` method was
actually broken. The `execute()` method returns an `ArrayList<String>`,
so the cast to `String` was erroring.

Changed the cast to the appriopriate one and fixed the loop that parses
the individual call log entries. Added extra error handling for the
class cast exception.


asm89:
fixes #855

Issue #365, opened 2014-04-04
GET /session/:sessionId/log/types throws exception


csnover:
selendroid 0.9.0 on Android 19 (4.4)

Attempting to retrieve log types throws an exception.

```
INFO: return value from selendroid android server: 
org.json.JSONException: A JSONArray text must start with '[' at character 1
	at org.json.JSONTokener.syntaxError(JSONTokener.java:410)
	at org.json.JSONArray.<init>(JSONArray.java:113)
	at org.json.JSONArray.<init>(JSONArray.java:157)
	at io.selendroid.server.handler.GetLogTypes.handle(GetLogTypes.java:31)
	at io.selendroid.server.SelendroidServlet.handleRequest(SelendroidServlet.java:140)
	at io.selendroid.server.BaseServlet.handleHttpRequest(BaseServlet.java:70)
	at org.webbitserver.netty.NettyHttpControl.nextHandler(NettyHttpControl.java:78)
	at org.webbitserver.netty.NettyHttpControl.nextHandler(NettyHttpControl.java:62)
	at org.webbitserver.handler.PathMatchHandler.handleHttpRequest(PathMatchHandler.java:33)
	at org.webbitserver.netty.NettyHttpControl.nextHandler(NettyHttpControl.java:78)
	at org.webbitserver.netty.NettyHttpControl.nextHandler(NettyHttpControl.java:62)
	at org.webbitserver.handler.DateHeaderHandler.handleHttpRequest(DateHeaderHandler.java:21)
	at org.webbitserver.netty.NettyHttpControl.nextHandler(NettyHttpControl.java:78)
	at org.webbitserver.netty.NettyHttpControl.nextHandler(NettyHttpControl.java:62)
	at org.webbitserver.handler.ServerHeaderHandler.handleHttpRequest(ServerHeaderHandler.java:25)
	at org.webbitserver.netty.NettyHttpControl.nextHandler(NettyHttpControl.java:78)
	at org.webbitserver.netty.NettyHttpControl.nextHandler(NettyHttpControl.java:67)
	at org.webbitserver.netty.NettyHttpChannelHandler$2.run(NettyHttpChannelHandler.java:72)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
	at java.lang.Thread.run(Thread.java:744)
```

[...]

PR #927, opened 2015-06-29
[standalone] Fix GetLogTypes handler

asm89:
Summary: The constructor of `JSONArray` that takes a string assumes that
the string is a JSON document. In the handler we want to construct a new
JSONArray to serialize it.

Test Plan:
Start selendroid-standalone and curl the `.../log/types` endpoint.
```
$ java -jar selendroid-standalone/target/selendroid-standalone-0.16.0-SNAPSHOT-with-dependencies.jar
$ $ curl -v localhost:4444/wd/hub/session/id/log/types
...
{"status":0,"sessionId":"id","value":["logcat"]}
```


asm89:
Fixes #365.

PR #928, opened 2015-06-29
[standalone] Fix HTTP response for unknown commands

asm89:
Summary: Before this fix the selendroid standalone server would proxy
any command to the server on the device. The server on the device would
return `404 Not Found`. The standalone server then failed to deserialize
that string as JSON and actually ended up retrying proxying the command
to the device three times.

Test Plan: Create a new session and try to query existing and
non-existing endpoints:

```
$ curl localhost:4444/wd/hub/session/8fbdf5e5-bc39-7012-09f8-0417acdf6916/foo
{"status":0,"sessionId":"8fbdf5e5-bc39-7012-09f8-0417acdf6916","value":"UNKNOWN_COMMAND"}
$ curl localhost:4444/wd/hub/session/8fbdf5e5-bc39-7012-09f8-0417acdf6916/
{"status":0,"sessionId":"8fbdf5e5-bc39-7012-09f8-0417acdf6916",...}
$ curl -XPOST localhost:4444/wd/hub/session/8fbdf5e5-bc39-7012-09f8-0417acdf6916/log
{"status":0,"sessionId":"8fbdf5e5-bc39-7012-09f8-0417acdf6916","value":["[2015-0...}
```

### DIFF
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/server/handler/ProxyToDeviceHandler.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/server/handler/ProxyToDeviceHandler.java
@@ -83,6 +83,9 @@ public class ProxyToDeviceHandler extends BaseSelendroidStandaloneHandler {
     while (retries-- > 0) {
       try {
         response = proxyRequestToDevice(request, session, url, method);
+        if (response == null) { // Unknown command
+          return new SelendroidResponse(sessionId, StatusCode.UNKNOWN_COMMAND);
+        }
         break;
       } catch (Exception e) {
         if (retries == 0) {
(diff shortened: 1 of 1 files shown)


selendroid-standalone/src/main/java/io/selendroid/standalone/server/handler/ProxyToDeviceHandler.java
```diff
@@ -83,6 +83,9 @@ public Response handleRequest(HttpRequest request, JSONObject payload) throws JS
     while (retries-- > 0) {
       try {
         response = proxyRequestToDevice(request, session, url, method);
+        if (response == null) { // Unknown command
+          return new SelendroidResponse(sessionId, StatusCode.UNKNOWN_COMMAND);
```
lukeis: shouldn't we be sending back the 404 to the client too? and not just wrapping the 404 from the server?

asm89: Yes, we should, but most clients should handle this version fine as well if I'm correct. At least this version doesn't throw some serialized exception at you.

Returning a 404 will require more work [as the code that handles the `SelendroidResponse` down the stack](https://github.com/selendroid/selendroid/blob/97e063cd2a28db269b5049848d2e32fba5e6e4d9/selendroid-server-common/src/main/java/io/selendroid/server/common/BaseServlet.java#L123-L136) is not aware of status codes based on the `SelendroidResponse` at all yet.

If you're ok with it I propose we merge this PR (given that clients will handle it) and I'll create a new issue to fix returning the right HTTP status code as well.

PR #929, opened 2015-06-30
[server] Fix NPE in ViewHierarchyAnalyzer

asm89:
Summary: We are seeing cases where we're querying native elements, but
there is no current activity. Previously this would result in a
`NullPointerException` in the `ViewHierarchyAnalyzer`. This changes
fixes the exception and adds a fallback when looking for resources:

- added a check to see if the activity is `null`
- if that's the case, try to get the resources from the passed view
- before getting trying to get the resource name, check if the resources
  returned are non null (we've seen that case as well)

Test Plan: Ran our internal e2e tests against selendroid with this
patch.


lukeis:
looks good to me, feel free to merge after travis comes back green

asm89:
More green! :tada:

Issue #800, opened 2015-02-03
Selendroid searches in top level views that are not visible


asm89:
In some cases selendroid ends up looking for elements in multiple top level views. I have a few questions about that:

1) What are valid cases where selendroid should look at multiple top level views?
2) Currently selendroid's `searchElements()` will also search for elements in views that are not the current root view. Is that intended behavior or a bug?

It seems that it's unintended because the element cache will never return elements that are not part of the current root view: [KnowsElement#get()](https://github.com/selendroid/selendroid/blob/eb50a4474c248f0610212830a828d2f1d65f1b20/selendroid-server/src/main/java/io/selendroid/server/model/KnownElements.java#L48). That means selendroid should also never have to search in those views?

These questions came to mind while working on #799. The proposed patch fixes a case for us where we would look up an element by id, but the element was present in multiple top level views that selendroid ended up searching. The way the search happened before would return the element that was not a child of the current root view, thus failing the following check when retrieving the element from cache when clicking: [KnowsElement#get()](https://github.com/selendroid/selendroid/blob/eb50a4474c248f0610212830a828d2f1d65f1b20/selendroid-server/src/main/java/io/selendroid/server/model/KnownElements.java#L48). Without changing the behavior of selendroid we are not able to make this test pass it seems.

aluedeke:
is this issue fixed in master or not?

asm89:
@aluedeke No, I'm awaiting response from @DominikDary as to why the searching in multiple top level views exists in the first place.

aluedeke:
@asm89 @DominikDary any updates? without having that fixed Selendroid is super unstable

DominikDary:
For me it also looks like that it is unintended.
I added the element cache verification that will never return elements that are not part of the current root view. So therefore I guess it would be the best that selendroid should  never have to search in those views.

asm89:
@DominikDary thanks for responding. I'll create a patch that fixes the behavior then.

DominikDary:
@asm89 thanks!

asm89:
Fixed by https://github.com/selendroid/selendroid/commit/85d3700a16a2fa76c1aa7bc2f0c327d1452d5ac4

aluedeke:
finally thx @asm89

PR #931, opened 2015-06-30
[server] Do not search in multiple top level views

asm89:
Summary: In some cases selendroid ended up looking for elements in
multiple top level views. This resulted in cases where we would look up
an element by id, but the element was present in in a top level view
that was not a child of the current root view. The server happily
returns the element, but as soon as you try to interact with the
element, the known elements cache will check if the element is part of
the root level view and if that's not the case throw a
`StaleElementReference` without further information.

Instead of this confusing behavior, selendroid should search only in the
current top level view.

Test Plan:
- Ran our internal e2e tests against selendroid with this patch.
- unit tests

### DIFF
--- a/selendroid-server/src/main/java/io/selendroid/server/model/DefaultSelendroidDriver.java
+++ b/selendroid-server/src/main/java/io/selendroid/server/model/DefaultSelendroidDriver.java
@@ -438,6 +438,11 @@ public class DefaultSelendroidDriver implements SelendroidDriver {
     }
 
     @Override
+    protected View getSearchRoot() {
+      return viewAnalyzer.getRecentDecorView();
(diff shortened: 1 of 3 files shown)


asm89:
For more context see #800 which is fixed by this PR.

asm89:
Note: I'm not too happy what's going on with the `getTopLevelViews()`, `getRecentDecorView()` etc stuff, but we also have a potential change coming up cleaning up the code that's responsible for searching for views etc. I'd like to defer cleaning these things up until that fix.

lukeis:
looks good to me, feel free to merge after travis comes back green

asm89:
Green! :tada:

PR #941, opened 2015-07-13
refactor By#findElement(s): save on creating new instances

alb-i986:
When forwarding to SearchContext, don't create a new copy of this; pass `this` instead.

PS: I'm very much new to this codebase, so please bear with me if I made any mistakes.
Thanks


alb-i986:
..I have signed the CLA.

PR #949, opened 2015-07-27
Extract method that converts BufferedImage to byte array

frolovs:


asm89:
lgtm :+1: 

cc @lukeis

PR #962, opened 2015-08-14
add more node options for grid auto-register

uchida:
This pull request adds [selenium grid node options]( https://github.com/SeleniumHQ/selenium/blob/master/java/server/src/org/openqa/grid/common/defaults/NodeOptions.json) support to selendroid grid auto-register.
These help to improve the selendroid scaling and high availability.

The default option value for these options are configured to fit selenium-server default ones with exception of `browserTimeout`.
I put the default `browserTimeout` value should be equal to `sessionTimeoutSeconds`, its default value is 30min.

### DIFF
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/SelendroidConfiguration.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/SelendroidConfiguration.java
@@ -100,6 +100,31 @@ public class SelendroidConfiguration {
           ".Allow to restart the hub without having to restart the nodes. 0 to disable auto register. Default 0.")
   private long registerCycle = 0;
 
+  @Parameter(names = "-cleanupCycle",
+          description = "How often in ms a proxy will check for timed out thread. Default 5000")
+  private long cleanupCycle = 5000;
+
+  @Parameter(names = "-timeout", description = "The timeout in seconds before the hub automatically ends a test " +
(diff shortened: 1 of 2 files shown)


uchida:
I have pushed additional commit.
This correct number of maxInstances and maxSession for grid auto-register.
With current implementation of selendroid, it seems impossible to run multiple tests on single device at a time, so maxInstance should be 1 and putting number of devices into maxSession is preferable.
This will keep selendroid behavior of existing these options because this changes only their default values.

selendroid-standalone/src/main/java/io/selendroid/standalone/SelendroidConfiguration.java
```diff
@@ -91,15 +91,40 @@
   private boolean keepAdbAlive = false;
 
   @Parameter(names = "-maxSession", description = "Maximum number of sessions that a grid hub can assign at a time.")
-  private int maxSession = 5;
+  private int maxSession = 0;
```
lukeis: i don't think it makes sense to set the default here to 0

lukeis: ah, i see it defaults to the devices available... maybe might want to change the description?

uchida:
For @lukeis comment, I added description about `-maxSession` default value.

lukeis:
thanks!

PR #964, opened 2015-08-16
modify regular expression to find platform folder for preview version

uchida:
Currently build tool for Android M preview is available and this cause the error `Command 'aapt' was not found inside the Android SDK. Please update to the latest development tools and try again.`.
This pull request evade this error.


lukeis:
the M release is no longer preview, so this shouldn't be an issue anymore.. Not sure if we should predict the next version to be preview released or not? Also, the issue could have been solved by installing an older build tools from the android sdk.

uchida:
Right, close this issue.

PR #965, opened 2015-08-17
 - Refactor selendroid-standlone to be able to create a server in a r…

phyllipecesar:
…emote host instead of always falling back to the localhost.

 - Abstract Emulator that implements both AndroidEmulator and AndroidDevice, so other people can have their own emulator version.
 - Bug fixes
   - Instead of accessing directly the targetPlatform variable, use the get method instead.


selendroid-standalone/src/main/java/io/selendroid/standalone/SelendroidConfiguration.java
```diff
@@ -143,6 +143,9 @@
   @Parameter(names = "-deleteTmpFiles", description = "Deletes temporary files created by the Selendroid Server.")
   private boolean deleteTmpFiles = true;
 
+  @Parameter(names = "-hostname", description = "The hostname that we should start the selendroid server")
+  private String hostname = "localhost";
```
asm89: `emulatorHostname`?

selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/DefaultAndroidEmulator.java
```diff
@@ -190,63 +167,33 @@ public static Dimension getScreenSizeFromSkin(String skinName) {
...
-          String avdName = client.sendCommand("avd name");
-          mapping.put(avdName, port);
-        } catch (AndroidDeviceException e) {
-          // ignore
-        } finally {
-          if (client != null) {
-            client.close();
-          }
-        }
+        String avdName = "test_remote_test";
```
asm89: Can you revert this and make sure it still works?

selendroid-standalone/src/main/java/io/selendroid/standalone/server/SelendroidStandaloneServer.java
```diff
@@ -45,9 +47,10 @@ protected SelendroidStandaloneServer(SelendroidConfiguration config,
     init();
   }
 
-  public SelendroidStandaloneServer(SelendroidConfiguration config)
+  public SelendroidStandaloneServer(SelendroidConfiguration config, InitAndroidDevicesStrategy initAndroidDevicesStrategy)
```
asm89: Perhaps have a constructor that only takes config and defaults to the old behavior.

selendroid-standalone/src/main/java/io/selendroid/standalone/server/model/AbstractDeviceStore.java
```diff
@@ -0,0 +1,197 @@
...
+import io.selendroid.standalone.android.AndroidEmulator;
+import io.selendroid.standalone.android.AndroidEmulatorPowerStateListener;
+import io.selendroid.standalone.android.impl.AbstractAndroidDeviceEmulator;
+import io.selendroid.standalone.exceptions.AndroidDeviceException;
+import io.selendroid.standalone.exceptions.DeviceStoreException;
+
+import java.util.*;
+import java.util.logging.Logger;
+
+public abstract class AbstractDeviceStore implements DeviceStore {
```
asm89: Did you change any code here or is it just a move/copy? Asking because I do have some comments on parts of the code, but if you didn't change it I'd rather keep it as is for now.

selendroid-standalone/src/test/java/io/selendroid/standalone/server/model/SelendroidStandaloneDriverTest.java
```diff
@@ -103,30 +120,12 @@ protected void assertThatTestAppHasBeenSuccessfullyRegistered(SelendroidStandalo
...
-    createTestSession(conf, caps);
-  }
-
-  @Test
-  public void shouldCreateNewTestSessionIfNoApkPassed() throws Exception {
-    SelendroidCapabilities caps = new SelendroidCapabilities();
-    caps.setAut(TEST_APP_ID);
-    caps.setLaunchActivity(TEST_APP_LAUNCH_ACTIVITY);
-    createTestSession(new SelendroidConfiguration(), caps);
-  }
```
asm89: Revert this (you moved the test methods?).

selendroid-standalone/src/main/java/io/selendroid/standalone/server/util/HttpClientUtil.java
```diff
@@ -41,12 +41,12 @@ public static HttpClient getHttpClient() {
   }
 
   public static HttpResponse executeRequestWithPayload(String uri, int port, HttpMethod method,
```
asm89: Wondering if we should have an overload where you don't pass the hostname that defaults to `localhost`. People might be using this in tests (which they probably shouldn't, but still..).

asm89:
cc @lukeis @DominikDary

selendroid-standalone/src/main/java/io/selendroid/standalone/server/model/AbstractDeviceStore.java
phyllipecesar: Just copy and paste

selendroid-standalone/src/test/java/io/selendroid/standalone/server/model/SelendroidStandaloneDriverTest.java
phyllipecesar: Yes, moved the test to the top, reverting...

PR #968, opened 2015-08-25
Branch fix

phyllipecesar:
Attaching asm suggestions!


asm89:
cc @lukeis @DominikDary

lukeis:
I'm confused about the use case for this. Why wouldn't that remote host be running the selendroid server? How would one use this?

asm89:
We're reconsidering this approach. In our case the remote host doesn't run the selendroid standalone, but we hook up already running emulators to selendroid standalone.

PR #975, opened 2015-09-11
add factory to allow custom drivers

lnicco:


asm89:
:+1: Pretty trivial.

cc @lukeis

PR #984, opened 2015-10-08
WebView needs to proxy WebViewClient & WebChromeClient

lukeis:
If the AUT is setting their own custom WebViewClient or WebChromeClient, we need to wrap it and call the AUT's object instead of overriding it.

See commit message and code comments for more info.

### DIFF
--- a/selendroid-server/src/main/java/io/selendroid/server/model/SelendroidWebDriver.java
+++ b/selendroid-server/src/main/java/io/selendroid/server/model/SelendroidWebDriver.java
@@ -380,15 +380,7 @@ public class SelendroidWebDriver {
             SelendroidLogger.info("Other WebView type, using SelendroidWebChromeClient");
             // get the existing WebChromeClient
             try {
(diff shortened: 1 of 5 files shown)


lukeis:
@asm89

selendroid-server/src/main/java/io/selendroid/server/model/SelendroidWebDriver.java
```diff
@@ -346,34 +367,65 @@ public void run() {
...
           // set handlers to indicate whether a page has started/done loading
-          view.setWebViewClient(new WebViewClient() {
-            @Override
-            public void onPageStarted(WebView view, String url, Bitmap favicon) {
-              synchronized (syncObject) {
-                pageStartedLoading = true;
-                syncObject.notify();
-              }
+          try {
+            Field f = view.getClass().getDeclaredField("mProvider");
```
JasnoWa: DRY this up? Looks like similar code to line 383-391

asm89:
:+1:

PR #985, opened 2015-10-13
wat the wrap.

lukeis:
Need to wrap the cordova clients as well.

My test app also likes to do "special" things to the cordova webview class.
The WebView's class name is that of a nested class and doesn't return anything for getSimpleName,
so need to rely on getting a ClassCast exception to determine if it really is cordova or not.


lukeis:
@asm89

asm89:
@lukeis :+1: go for it

PR #991, opened 2015-10-20
may need to run on UI thread to get attribute

lukeis:
This is needed especially if checking for attributes on a WebView

(in particular wanting to check WebView.getScale() )

@asm89


asm89:
:+1: I trust you on this one. Looks tricky though. :)

PR #998, opened 2015-10-27
Gradle build system

vishals79:
# This will enable a new build system.

**Changes**
1. Add root build gradle file along-with sub-projects gradle files.
2. Ignore 5 test.
3. Modify syntax for two test case.

**Build Command**
1. gradle clean build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: android
 android:
   components:
-    - build-tools-19.0.3
+    - build-tools-23.0.2
     - extra-google-m2repository
     - extra-android-m2repository
     - android-16
@@ -23,16 +23,28 @@ branches:
   only:
     - master
 
-before_install: 
+before_install:
+  # Comment gradle build command to use maven and vice versa
+  - wget http://services.gradle.org/distributions/gradle-2.7-bin.zip
+  - unzip -qq gradle-2.7-bin.zip
+  - export GRADLE_HOME=$PWD/gradle-2.7
+  - export PATH=$GRADLE_HOME/bin:$PATH
   - "git clone -b travis `git config --get remote.origin.url` target/travis"
-  - "mvn --version"
-  - "cd third-party/"
-  - "./mvnInstall.sh"
-  - "cd ../"
+
+  # Maven
+  # - "mvn --version"
+  # - "cd third-party/"
+  # - "./mvnInstall.sh"
+  # - "cd ../"
+
+before_script:
+  - echo no | android create avd --force -n l10n -t android-16 --abi armeabi-v7a
 
 script: 
-  - "[ ${TRAVIS_PULL_REQUEST} = 'false' ] && mvn clean deploy --settings target/travis/settings.xml || mvn clean verify --settings target/travis/settings.xml"
+  - "gradle clean build"
 
+  # Maven
+  #  - "[ ${TRAVIS_PULL_REQUEST} = 'false' ] && mvn clean deploy --settings target/travis/settings.xml || mvn clean verify --settings target/travis/settings.xml"
 
 env:
   global:
--- a/selendroid-common/build.gradle
+++ b/selendroid-common/build.gradle
@@ -1,0 +1,17 @@
+apply plugin: 'java'
+
+dependencies {
+    compile rootProject.ext.json
+    compile rootProject.ext.junit
+    // https://bintray.com/bintray/jcenter/org.seleniumhq.selenium%3Aselenium-java/2.48.2/view#files
+    compile("org.seleniumhq.selenium:selenium-java:$rootProject.ext.seleniumVersion") {
+        exclude(module: 'selenium-chrome-driver')
+        exclude(module: 'selenium-edge-driver')
+        exclude(module: 'selenium-htmlunit-driver')
+        exclude(module: 'selenium-firefox-driver')
+        exclude(module: 'selenium-ie-driver')
+        exclude(module: 'selenium-safari-driver')
+        exclude(module: 'webbit')
+        exclude(module: 'selenium-leg-rc')
+    }
+}
--- a/selendroid-grid-plugin/build.gradle
+++ b/selendroid-grid-plugin/build.gradle
@@ -1,0 +1,5 @@
+apply plugin: 'java'
+
+dependencies {
+    compile("org.seleniumhq.selenium:selenium-server:$rootProject.ext.seleniumVersion")
+}
--- a/selendroid-server/build.gradle
+++ b/selendroid-server/build.gradle
@@ -1,0 +1,75 @@
+buildscript {
+    repositories {
+        jcenter()
+    }
+    dependencies {
+        classpath rootProject.ext.androidBuildToolsGradle
+        classpath rootProject.ext.sdkManager
+    }
+}
+
+apply plugin: 'android-sdk-manager'
+apply plugin: 'com.android.application'
+
+android {
+    compileSdkVersion rootProject.ext.compileSdkVersion
+    buildToolsVersion rootProject.ext.buildToolsVersion
+
+    defaultConfig {
+        minSdkVersion rootProject.ext.minSdkVersion
+        targetSdkVersion rootProject.ext.targetSdkVersion
+    }
+
+    lintOptions {
+        abortOnError false
+    }
+
+    sourceSets {
+        main {
+            manifest.srcFile 'AndroidManifest.xml'
+            res.srcDirs = ['res']
+            assets.srcDirs = ['assets']
+        }
+    }
+
+    packagingOptions {
+        exclude 'META-INF/INDEX.LIST'
+        exclude 'META-INF/io.netty.versions.properties'
+    }
+
+    buildTypes {
+        debug {
+            applicationVariants.all { variant ->
+                variant.outputs.each { output ->
+                    output.outputFile = new File(
+                            output.outputFile.parent,
+                            output.outputFile.name.replace('-debug.apk', "-${variant.versionName}.apk"))
+                }
+            }
+        }
+    }
+}
+
+configurations {
+    provided
+}
+
+sourceSets {
+    main.compileClasspath += configurations.provided
+    test.compileClasspath += configurations.provided
+    test.runtimeClasspath += configurations.provided
+}
+
+dependencies {
+    compile project(':selendroid-server-common')
+    compile 'com.android.support:support-v4:23.1.0'
+    testCompile rootProject.ext.json
+    testCompile rootProject.ext.junit
+    testCompile 'org.mockito:mockito-core:1.9.5'
+    testCompile 'commons-io:commons-io:2.2'
+    testCompile 'org.apache.httpcomponents:httpclient:4.2.1'
+    testCompile 'commons-logging:commons-logging:1.1.3'
+
+    provided files('../third-party/cordova-3.7.0/classes.jar')
+    provided files('../third-party/cordova-4.0.0/classes.jar')
+}
--- a/selendroid-standalone/build.gradle
+++ b/selendroid-standalone/build.gradle
@@ -1,0 +1,59 @@
+apply plugin: 'java'
+apply plugin: 'application'
+
+dependencies {
+    compile project(':selendroid-common')
+    compile project(':selendroid-server-common')
+    compile 'org.apache.httpcomponents:httpclient:4.3.4'
+    compile rootProject.ext.json
+    compile 'commons-io:commons-io:2.2'
+    compile 'org.apache.commons:commons-exec:1.1'
+    compile 'org.apache.commons:commons-compress:1.5'
+    compile 'com.beust:jcommander:1.30'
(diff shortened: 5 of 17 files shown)


.travis.yml
```diff
@@ -25,14 +25,31 @@ branches:
...
+  - "gradle clean build"
+  - "cd .."
+  - "cd selendroid-server-common"
+  - "gradle clean build"
+  - "cd .."
+  - "cd selendroid-standalone"
+  - "gradle clean build"
+  - "cd .."
+  - "cd selendroid-test-app"
+  - "gradle assembleDebug"
```
lukeis: can you define a [multi-project gradle build file](https://docs.gradle.org/current/userguide/multi_project_builds.html) at the root level and then we can just do `gradle build` from the root of the project to kick off all of these? (and somehow map that test-app needs to call assembleDebug when we 'build' from the root)

lukeis: also, the maven verify target will execute the tests, so we should add the gradle command to run the tests too.

selendroid-common/build.gradle
```diff
@@ -0,0 +1,18 @@
+apply plugin: 'java'
+
+repositories {
+    jcenter()
+}
+
+dependencies {
+    compile 'org.json:json:20090211'
+    compile 'junit:junit:4.8.2'
+    compile('org.seleniumhq.selenium:selenium-java:2.47.2') {
```
lukeis: bump to 2.48.2

selendroid-grid-plugin/build.gradle
```diff
@@ -0,0 +1,9 @@
+apply plugin: 'java'
+
+repositories {
+    jcenter()
+}
+
+dependencies {
+    compile 'org.seleniumhq.selenium:selenium-server:2.45.0'
```
lukeis: bump to 2.48.2  ... is there a way to define the selenium server version is a 'root' gradle settings file so we don't have to duplicate it everywhere?

selendroid-server/build.gradle
```diff
@@ -0,0 +1,81 @@
...
+    testCompile 'org.json:json:20090211'
+    testCompile 'junit:junit:4.12'
+    testCompile 'org.mockito:mockito-core:1.9.5'
+    testCompile 'commons-io:commons-io:2.2'
+    testCompile 'org.apache.httpcomponents:httpclient:4.2.1'
+    testCompile 'commons-logging:commons-logging:1.1.3'
+
+    provided files('../third-party/cordova-3.7.0/classes.jar')
+    provided files('../third-party/cordova-4.0.0/classes.jar')
+    provided files('../third-party/support-v4/classes.jar')
```
lukeis: this support classes.jar should be available in the android library in the sdk, we should be able to remove it from third-party (well, we can remove it when we remove the pom files)

selendroid-standalone/build.gradle
```diff
@@ -0,0 +1,57 @@
...
+
+    // Version from manifest attributes will be retrieved while building selendroid-server apk
+    manifest {
+        attributes('Main-Class': "$mainClassName", 'version': '0.18.0-SNAPSHOT')
+    }
+
+    into('prebuild') {
+        from {
+            ['../selendroid-server/build/outputs/apk/selendroid-server-0.18.0-SNAPSHOT.apk',
+             '../android-driver/build/outputs/apk/android-driver-app-0.18.0-SNAPSHOT.apk']
```
lukeis: 0.18 seems to be hard coded in a few different places, can we put that in a root level settings file?

selendroid-standalone/src/test/java/io/selendroid/standalone/android/impl/DefaultHardwareDeviceTests.java
```diff
@@ -61,6 +62,7 @@ private void cleanUpDevice(AndroidDevice emulator) throws Exception {
     Assert.assertFalse(installedAPKs.contains(AUT_PACKAGE));
   }
 
+  @Ignore("Fail. Requires an active emulator to install the selendroid apk.")
```
lukeis: something doesn't seem right here... this test mocks everything, it should not require an active emulator or device.

selendroid-standalone/src/test/java/io/selendroid/standalone/android/impl/DefaultAndroidEmulatorTests.java
```diff
@@ -23,15 +23,19 @@
 import org.apache.commons.io.FileUtils;
 import org.junit.Assert;
 import org.junit.Test;
+import org.junit.Ignore;
 import org.openqa.selenium.Dimension;
 
 public class DefaultAndroidEmulatorTests {
+
+  @Ignore("Fail. Requires list of available avds")
```
lukeis: this should only fail if you have not defined any avd's on your system? if you create one, should be ok?

selendroid-standalone/src/test/java/io/selendroid/standalone/android/impl/DefaultAndroidEmulatorTests.java
```diff
@@ -23,15 +23,19 @@
...
   @Test
   public void shouldBeAbleToListAvds() throws Exception {
     List<AndroidEmulator> avds = DefaultAndroidEmulator.listAvailableAvds();
     Assert.assertFalse("Expecting list of avds not to be empty", avds.isEmpty());
   }
 
+  @Ignore("Fail. Requires emulator to exist on the system running the test.")
```
lukeis: are we assuming a system that is running these tests does not have any emulators? or can't create one?

selendroid-standalone/src/test/java/io/selendroid/standalone/android/impl/AbstractDeviceTest.java
```diff
@@ -20,6 +20,7 @@
 import org.hamcrest.Description;
 import org.hamcrest.Matcher;
 import org.junit.Test;
+import org.junit.Ignore;
```
lukeis: unnecessary import?

selendroid-standalone/src/test/java/io/selendroid/standalone/android/impl/DefaultHardwareDeviceTests.java
bootstraponline: the comment in this test file says `Works only if ... one emulator is already running`

selendroid-standalone/src/test/java/io/selendroid/standalone/android/impl/DefaultAndroidEmulatorTests.java
bootstraponline: It's failing on Travis which I'd expect to already have an avd setup? Otherwise the test wouldn't pass under maven.

selendroid-server/build.gradle
bootstraponline: nice catch, it should be:

`'com.android.support:support-v4:23.0.1'`

DominikDary:
Thanks for providing your pull request!

When I look at the travis file, the build of the whole project cannot be triggered from the root folder, right?
I think it is very important to have this single point of entry.

vishals79:
@DominikDary : Thanks for you response. 
We are updating the project to create a root build system and will very soon update the PR.

.travis.yml
vishals79: Added multi-project root build.
Using _gradle clean build_ at the root will build the complete project and execute tests (tests skipped for selendroid-test-app)

selendroid-common/build.gradle
vishals79: Updated to 2.48.2

selendroid-grid-plugin/build.gradle
vishals79: Updated to 2.48.2 and added the version number at the root level.

selendroid-server/build.gradle
vishals79: Removed the third party library for support-v4 and addded 'com.android.support:support-v4:23.1.0' as a dependency.

selendroid-standalone/build.gradle
vishals79: Added version = '0.18.0-SNAPSHOT' at the root level.

selendroid-standalone/src/test/java/io/selendroid/standalone/android/impl/DefaultHardwareDeviceTests.java
vishals79: This test is ignored as we don't have one running emulator.

selendroid-standalone/src/test/java/io/selendroid/standalone/android/impl/DefaultAndroidEmulatorTests.java
vishals79: Fixed. Created one avd on Travis with details
android create avd --force -n l10n -t android-16 --abi armeabi-v7a

selendroid-standalone/src/test/java/io/selendroid/standalone/android/impl/DefaultAndroidEmulatorTests.java
vishals79: This test is still ignored as the test is unable to start the emulator.

selendroid-standalone/src/test/java/io/selendroid/standalone/android/impl/AbstractDeviceTest.java
vishals79: Removed the unnecessary import.

lukeis:
:+1: 

@asm89 @DominikDary please review

DominikDary:
Thanks @vishals79 !
I just had another question: If we would like to push to maven central, how can we do that?

lukeis:
@DominikDary this seems to have the info necessary:

http://jedicoder.blogspot.com/2011/11/automated-gradle-project-deployment-to.html

vishals79:
@DominikDary @lukeis : We have one request raised by @bootstraponline for publishing of jars to jcenter and maven. 
https://github.com/appium/selendroid/issues/2

bootstraponline:
> I just had another question: If we would like to push to maven central, how can we do that?

Gradle uses jcenter by default. Publishing to maven central alone is insufficient. What I'm proposing for selendroid is:

- [jfrog artifactory](https://github.com/appium/droiddriver/blob/72a40802f8e7d24a3175e09adc4b44a89df54f38/artifactory.gradle) for nightly snapshots
- [maven plugin](https://github.com/appium/droiddriver/blob/72a40802f8e7d24a3175e09adc4b44a89df54f38/maven.gradle) for pom generation
- [bintray jcenter](https://github.com/appium/droiddriver/blob/master/jcenter.gradle) for publishing artifacts to both jcenter and [maven central](http://blog.bintray.com/2014/02/11/bintray-as-pain-free-gateway-to-maven-central/) via [synching](https://bintray.com/docs/usermanual/uploads/uploads_syncingartifactswithmavencentral.html)

I've setup this for droiddriver already on appium's cloudbees Jenkins. I think the same approach would work for selendroid + travis. In general I find Jenkins more flexible though.

I think the next step is for the imaginea team to update the PR with support for publishing to jcenter/maven central (assuming selendroid committers agree with this plan).

DominikDary:
The current setup is snapshots are pushed to sonatype.org only and binaries are pushed only to maven central.

For me it is key to support this mainly. If we sync somewhere else, fine, but that should not be the priority.

@asm @lukeis any thougts?

bootstraponline:
> The current setup is snapshots are pushed to sonatype.org only and binaries are pushed only to maven central.
> 
> For me it is key to support this mainly. If we sync somewhere else, fine, but that should not be the priority.

jcenter is what Google uses by default in Android Studio so that's a requirement. There's a [sonatype](https://github.com/bmuschko/gradle-nexus-plugin) gradle plugin. If we publish releases to jcenter (which simultaneously publishes to maven central) and artifacts to sonatype then that should be sufficient.

bootstraponline:
Selendroid will need a jcenter organization along with maven central and sonatype credentials. Does someone from the selendroid team want to setup these accounts or should I do it?

For now we're going to validate the publishing code using the appium credentials. It should just be a matter of replacing the appium credentials with selendroid's once the accounts are setup.

Appium doesn't have a sonatype account so it's not clear how we're going to test that. It's painful to make an account there.

DominikDary:
I would like to setup an account.

DominikDary:
@lukeis and myself had a quick chat about the PR. We think we should merge it in. The reason is, it adds the graddle build system but does not yet remove the maven one, so there we are fine.

@vishals79 : can you please sign the CLA: https://docs.google.com/forms/d/1PwOm260Al4nIUTabt_Pj-962OzkPxHWvK9Pw5hy_eUM/viewform

Thanks

vishals79:
@DominikDary : Thanks! We have signed the CLA. In PR, we have gradle build configured in the travis.yml ( maven configurations are commented).

asm89:
:+1:

DominikDary:
Thanks fro signing the CLA!

PR #1011, opened 2015-12-01
Add option to start selendroid standalone on a random port

asm89:
There are cases where you don't care about the port selendroid
standalone starts on, as long as you can get the port after it started.
This commit changes the webserver start method to block until the netty
channel has opened and the netty thread returned the actual port the
server has started on.

Starting the standalone server on a random port can be achieved by using
0 as the configured port.


asm89:
cc @lukeis @DominikDary

PR #1012, opened 2015-12-01
Add option to re-use configured server port

asm89:
For some reason the standalone server will assign a new selendroid
server port to the client each time a new session is created. Made this
behavior configurable and default to whatever the behavior was before.


asm89:
cc @lukeis @DominikDary

DominikDary:
+1

PR #1055, opened 2016-05-22
Add support for launching implicit intents

bthood:
Sometimes a test may want to launch an Activity via URI (e.g. to encode
a user ID or other information). This adds an "intent_uri" option to
both lightweight and regular server instrumentation that acts as a
fallback in case the main activity name is not specified.

/cc @asm89


asm89:
👍

Issue #1062, opened 2016-06-20
Android N Support


harshagarwal8:
Unable to start driver in Android N device

lpetre:
This should be fixed by #1141, please get the latest code from master and try again.

PR #1063, opened 2016-06-20
Adding API Level for Android N

harshagarwal8:


operatiay:
does this fix your issue #1062?

lpetre:
Added in #1141

PR #1067, opened 2016-06-29
get ANDROID_HOME from Property

vesense:
If not found in OS  vars, lookup from JVM Properties.
This is very helpful for avoiding restart user's PC.


vesense:
`System.getenv(ANDROID_HOME)` will return `null` in Windows OS even if user specified the value. It dosen't works unless restarting the machine.

PR #1013, opened 2015-12-02
Adding Uiautomator V2 driver to selendroid Factory

anandsadu:
Tried Integrating Uiautomator 2 driver in to the selendroid factory by reusing the existing SelendroidDriver interface and ServerInstrumentation instance. The current driver is partially implemented to test one action ie 'pressMenu' on the device. But  While getting in the instance of the UIDevice an exception was thrown which was mentioned below.

Borrowed test case from the following link to run a sample test ie findElement.
https://github.com/selendroid/demoproject-selendroid/blob/8aa982a3b15218a2dccb63fb56ed8425d53f9b8f/src/main/java/io/selendroid/demo/SelendroidIntegrationTest.java

```
E/SELENDROID(30203): io.selendroid.server.handler.SafeRequestHandler#handle:110 - Exception while handling action in: io.selendroid.server.handler.FindElement
E/SELENDROID(30203): java.lang.NullPointerException: Attempt to invoke virtual method 'void android.app.UiAutomation.setOnAccessibilityEventListener(android.app.UiAutomation$OnAccessibilityEventListener)' on a null object reference
E/SELENDROID(30203):    at android.support.test.uiautomator.UiAutomatorBridge.setOnAccessibilityEventListener(UiAutomatorBridge.java:78)
E/SELENDROID(30203):    at android.support.test.uiautomator.QueryController.<init>(QueryController.java:58)
E/SELENDROID(30203):    at android.support.test.uiautomator.UiAutomatorBridge.<init>(UiAutomatorBridge.java:66)
E/SELENDROID(30203):    at android.support.test.uiautomator.InstrumentationUiAutomatorBridge.<init>(InstrumentationUiAutomatorBridge.java:35)
E/SELENDROID(30203):    at android.support.test.uiautomator.UiDevice.<init>(UiDevice.java:103)
E/SELENDROID(30203):    at android.support.t...
```

PR #1068, opened 2016-07-06
Enabled Selendroid to access the UiAutomation APIs

marcelogomez:
[UiAutomation](https://developer.android.com/reference/android/app/UiAutomation.html) is a low level API that allow us to inject events across package boundaries using the system's accessibility APIs. As of API level 18, an instance of it can be obtained via [Instrumentation.getUiAutomation](https://developer.android.com/reference/android/app/Instrumentation.html#getUiAutomation()), however, this method will fail unless the instrument command is run with the -w flag and it is called from a test thread.

This pull request refactors ServerInstrumentation to extend [AndroidJUnitRunner](https://developer.android.com/reference/android/support/test/runner/AndroidJUnitRunner.html) and moves the call to start the http server to ServerManager, a dummy test


marcelogomez:
This would also take care of #1013

DominikDary:
Hi @marcelogomez,
thanks for the PR! The build is failing. Can you please have a look?

marcelogomez:
Hi @DominikDary !
I'm not sure the issue is related to my changes, it's a dependency conflict with hamcrest. From the logs:
> :selendroid-server:prepareDebugUnitTestDependenciesConflict with dependency 'org.hamcrest:hamcrest-core'. Resolved versions for app (1.3) and test app (1.1) differ. See http://g.co/androidstudio/app-test-app-conflict for details.

marcelogomez:
Making the test app explicitly depend on JUnit again which in turn depends on Hamcrest fixed it

DominikDary:
I will have a look next weekend.

DominikDary:
cc @asm89 

@marcelogomez This pull request looks very similar to #1072. Is this one obsolete with the other one or do they need to be merged in a specific order?

marcelogomez:
Yup, this is obsolete now

PR #1078, opened 2016-07-28
Bug fixes

marcelogomez:
Introducing the new JUnitRunner instrumentation introduced some bugs, which have been fixed in this PR:

- JUnit should no longer constrained to a test dependency
- bootstrap methods weren't being called
- SelendroidInstrumentation needs to be explicitly registered with InstrumentationRegistry

This also fixes other unrelated issues:

- AbstractDeviceTest was breaking the maven build


marcelogomez:
gradle clean build is working fine locally :/

marcelogomez:
cc @asm89 could you take a look?

PR #1103, opened 2017-03-09
Pass arbitrary arguments to AUT via instr args

marcelogomez:
This PR adds the ability to pass arguments to the Application Under Test via instrumentation arguments. This is done by adding a JSONObject to the required capabilities, these are then sent to the AUT via instrumentation arguments, where they can be read as strings. This only supports flat objects

### DIFF
--- a/selendroid-common/src/main/java/io/selendroid/common/SelendroidCapabilities.java
+++ b/selendroid-common/src/main/java/io/selendroid/common/SelendroidCapabilities.java
@@ -61,7 +61,7 @@ public class SelendroidCapabilities extends DesiredCapabilities {
   public static final String LAUNCH_ACTIVITY = "launchActivity";
   public static final String SELENDROID_EXTENSIONS = "selendroidExtensions";
   public static final String BOOTSTRAP_CLASS_NAMES = "bootstrapClassNames";
-  public static final String EXTRA_ARGS = "extraArgs";
+  public static final String EXTRA_ARGS = "extraAUTArgs";
 
   public static final String USE_JUNIT_BOOTSTRAP = "useJUnitBootstrap";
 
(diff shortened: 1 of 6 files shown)


selendroid-server/src/main/java/io/selendroid/server/InstrumentationArguments.java
```diff
@@ -28,15 +32,22 @@
...
+    mainActivityClassName = arguments.getString(SelendroidArguments.MAIN_ACTIVITY);
+    intentUri = arguments.getString(SelendroidArguments.INTENT_URI);
+    intentAction = arguments.getString(SelendroidArguments.INTENT_ACTION);
+    serviceClassName = arguments.getString(SelendroidArguments.SERVICE);
+    loadExtensions = Boolean.parseBoolean(arguments.getString(SelendroidArguments.LOAD_EXTENSIONS));
+    bootstrapClassNames = arguments.getString(SelendroidArguments.BOOTSTRAP);
+    serverPort = arguments.getString(SelendroidArguments.SERVER_PORT);
+
+    for (String key : arguments.keySet()) {
+      if (!SelendroidArguments.KNOWN_ARGUMENTS.contains(key)) {
```
asm89: nit: want to log a warning if someone tries?

selendroid-common/src/main/java/io/selendroid/common/SelendroidCapabilities.java
```diff
@@ -58,6 +61,7 @@
   public static final String LAUNCH_ACTIVITY = "launchActivity";
   public static final String SELENDROID_EXTENSIONS = "selendroidExtensions";
   public static final String BOOTSTRAP_CLASS_NAMES = "bootstrapClassNames";
+  public static final String EXTRA_ARGS = "extraArgs";
```
asm89: Perhaps it would be more clear if we call this something like `extraAUTArgs`? It isn't directly clear to me where the args go.

PR #1106, opened 2017-03-14
Remove Jacoco from selendroid-server

marcelogomez:
CI build is breaking because it's unable to instrument `selendroid-server/build/intermediates/classes/debug/io/selendroid/server/model/js/AndroidAtoms.class `. This PR removes Jacoco form the selendroid-server module to prevent this


asm89:
👍 at least the build is green again

PR #1109, opened 2017-04-06
Implement /wd/hub/session/:sessionId/element/:id/rect endpoint for better compatibility with Selenium 3

mykola-mokhnach:
Selenium 3 introduced new mandatory element property called _rect_. This PR adds the properly to Selendroid

### DIFF
--- a/selendroid-server/src/main/java/io/selendroid/server/android/internal/Rectangle.java
+++ b/selendroid-server/src/main/java/io/selendroid/server/android/internal/Rectangle.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2012-2014 eBay Software Foundation and selendroid committers.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ * 
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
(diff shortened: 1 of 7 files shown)


selendroid-server/src/main/java/io/selendroid/server/android/internal/Rectangle.java
```diff
@@ -0,0 +1,67 @@
...
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package io.selendroid.server.android.internal;
+
+/**
+ * A copy of java.awt.Rectangle, to remove dependency on awt.
+ */
+public class Rectangle {
```
marcelogomez: I don't think we really need this class Android already has a Rect implementation
https://developer.android.com/reference/android/graphics/Rect.html

selendroid-server/src/main/java/io/selendroid/server/model/AndroidNativeElement.java
```diff
@@ -478,6 +479,12 @@ public Point getLocation() {
     return new Point(xy[0], xy[1]);
   }
 
+  @Override
+  public Rectangle getRect() {
+    Point location = getLocation();
```
marcelogomez: Can you take a look at the `click` implementation? Turns out that `getLocation` might not always work immediately :/

I think we can do some refactoring to make it safer

marcelogomez:
Thanks for taking a stab at this :)
I added a couple of comments

mykola-mokhnach:
@marcelogomez This PR was created more than half a year ago. It looks like the project is not maintained for quite a while. Do you plan to support it in the future? Are there better alternatives?

PR #1110, opened 2017-04-13
adding support for Android25

psubrownie:
I saw someone else's pull request for ANDROID24 so I left it out.  I also needed to update the ddmlib version to parse line numbers correctly.  using version 23 no properties were parsed.


lpetre:
ANDROID24 was fixed in #1141

I'm not sure what your ddmlib change does. Can you elaborate in a separate PR please?

PR #1117, opened 2017-06-08
Move extra AUT args to a JSON file

marcelogomez:
This PR changes the way in which we send extra AUT args. Instead of using the instrument command's arguments, we now write the arguments to a JSON file and push it to the device. This will free us from having to worry about the character limit on `adb` command arguments and will also allow us to send more structured args (we now support passing arbitrary `JSONObject`s)

### DIFF
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/server/model/SelendroidStandaloneDriver.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/server/model/SelendroidStandaloneDriver.java
@@ -1,11 +1,11 @@
 /*
  * Copyright 2012-2014 eBay Software Foundation and selendroid committers.
- * 
+ *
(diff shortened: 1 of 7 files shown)


selendroid-standalone/src/main/java/io/selendroid/standalone/server/model/SelendroidStandaloneDriver.java
```diff
@@ -358,6 +361,37 @@ private void waitForServerStart(AndroidDevice device) {
...
+  private void pushExtraArgsToDevice(
+    AndroidDevice device,
+    SelendroidCapabilities caps
+  ) {
+    String devicePath = new File(
+      device.getExternalStoragePath(),
+      "extra_args.json"
+    ).getAbsolutePath();
+
+    // Make sure we clean the file in case we don't push a new one
```
asm89: 👍

asm89:
LGTM

PR #1120, opened 2017-06-19
Add logging for instrumentation lifecycle methods

marcelogomez:
This PR does some refactoring of `DefaultServerInstrumentation` and `JUnitRunnerInstrumentation` to reduce redundancies and adds some log statements for the instrumentation's lifecycle

### DIFF
--- a/selendroid-server/src/main/java/io/selendroid/server/DefaultServerInstrumentation.java
+++ b/selendroid-server/src/main/java/io/selendroid/server/DefaultServerInstrumentation.java
@@ -60,6 +60,13 @@ public class DefaultServerInstrumentation implements ServerInstrumentation {
         this.instrumentation = instrumentation;
         this.args = args;
 
+        // Get a wake lock to stop the cpu going to sleep
+        PowerManager pm =(PowerManager) instrumentation
+          .getContext()
+          .getSystemService(Context.POWER_SERVICE);
+        this.wakeLock =
+          pm.newWakeLock(PowerManager.SCREEN_DIM_WAKE_LOCK, "Selendroid");
+
         if (args.isLoadExtensions()) {
             extensionLoader = new ExtensionLoader(instrumentation.getTargetContext(),
                     ExternalStorage.getExtensionDex().getAbsolutePath());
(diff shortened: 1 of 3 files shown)


selendroid-server/src/main/java/io/selendroid/server/DefaultServerInstrumentation.java
```diff
@@ -60,6 +60,13 @@ public DefaultServerInstrumentation(Instrumentation instrumentation,
         this.instrumentation = instrumentation;
         this.args = args;
 
+        // Get a wake lock to stop the cpu going to sleep
+        PowerManager pm =(PowerManager) instrumentation
```
asm89: nit: space

asm89: Should we be cautious here?

selendroid-server/src/main/java/io/selendroid/server/DefaultServerInstrumentation.java
```diff
@@ -146,12 +148,36 @@ public void startActivity(String activityClassName) {
 
     @Override
     public void startServer() {
+        SelendroidLogger.info("*** ServerInstrumentation#startSever() ***");
```
asm89: nit: `startServer`

selendroid-server/src/main/java/io/selendroid/server/DefaultServerInstrumentation.java
```diff
@@ -146,12 +148,36 @@ public void startActivity(String activityClassName) {
...
+                  SelendroidLogger.error("Failed to start Selendroid server", e);
+              }
+          }
+        });
+    }
+
+    // TODO: Make this class a base class and extract all the serverthread
+    // specific logic to a subclass
+    protected void startServerImpl() {
+        SelendroidLogger.info("*** ServerInstrumentation#startSeverImpl() ***");
```
asm89: s/Sever/Server

selendroid-server/src/main/java/io/selendroid/server/DefaultServerInstrumentation.java
marcelogomez: What do you mean? `null` check? I just moved this from somewhere else

asm89: Let's keep it like this for now then. Just caught my eye.

PR #1125, opened 2017-07-17
Increase click reliability for API levels 19+

marcelogomez:
The previous implementation of the click method would simply wait for a specific amount of time and then try to access the view's location. That approach offers no guarantee that the `getLocationOnScreen` would return a valid value.

This PR checks if the view's been laid out before calling `getLocationOnScreen`, if it hasn't been it attaches an `OnGlobalLayoutListener` to the view tree that will perform the click once the view's been properly laid out.

Note that `isLaidOut` is only available in API 19+, so we fallback to the original implementation for lower API levels

### DIFF
--- a/selendroid-server/src/main/java/io/selendroid/server/model/AndroidNativeElement.java
+++ b/selendroid-server/src/main/java/io/selendroid/server/model/AndroidNativeElement.java
@@ -259,13 +259,11 @@ public class AndroidNativeElement implements AndroidElement {
 
     final View view = getView();
 
-    final int viewWidth = view.getWidth();
-    final int viewHeight = view.getHeight();
     final int[] xy = new int[2];
 
     if (viewWidth > 0 && viewHeight > 0) {
       view.getLocationOnScreen(xy);
-      clickOnScreen(xy[0] + viewWidth / 2.0f, xy[1] + viewHeight / 2.0f);
+      clickOnScreen(xy[0] + view.getWidth() / 2.0f, xy[1] + view.getHeight() / 2.0f);
       return;
     }
 
@@ -275,7 +273,7 @@ public class AndroidNativeElement implements AndroidElement {
       public void onGlobalLayout() {
         try {
           view.getLocationOnScreen(xy);
-          clickOnScreen(xy[0] + viewWidth / 2.0f, xy[1] + viewHeight / 2.0f);
+          clickOnScreen(xy[0] + view.getWidth() / 2.0f, xy[1] + view.getHeight() / 2.0f);
         } finally {
           if (observer.isAlive()) {
             removeOnGlobalLayoutListener(observer, this);
--- a/selendroid-server/src/main/java/io/selendroid/server/model/AndroidNativeElement.java
+++ b/selendroid-server/src/main/java/io/selendroid/server/model/AndroidNativeElement.java
@@ -260,8 +260,6 @@ public class AndroidNativeElement implements AndroidElement {
 
     final View view = getView();
 
-    final int[] xy = new int[2];
-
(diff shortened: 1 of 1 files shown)


selendroid-server/src/main/java/io/selendroid/server/model/AndroidNativeElement.java
```diff
@@ -254,18 +256,73 @@ public String getText() {
...
+      if (Build.VERSION.SDK_INT < 19) {
+        try {
+          // This is needed for recalculation of location when we cannot
+          // explicitly check if the view's been laid out
+          Thread.sleep(300);
+        } catch (InterruptedException e) {}
+      }
+
+      view.getLocationOnScreen(xy);
+      clickOnScreen(xy[0] + view.getWidth() / 2.0f, xy[1] + view.getHeight() / 2.0f);
```
passy: Would it be useful to make those two lines a separate function as they're repeated twice and it seems possible to get out of sync if someone changes one without the other.

selendroid-server/src/main/java/io/selendroid/server/model/AndroidNativeElement.java
```diff
@@ -254,18 +256,73 @@ public String getText() {
   public void click() {
     waitUntilIsDisplayed();
     scrollIntoScreenIfNeeded();
+
+    final View view = getView();
+
+    final int[] xy = new int[2];
+
+    if (Build.VERSION.SDK_INT < 19 || view.isLaidOut()) {
```
passy: I actually didn't know about that call. That seems like a great way to reduce the flakiness of clicks.

selendroid-server/src/main/java/io/selendroid/server/model/AndroidNativeElement.java
```diff
@@ -254,18 +256,73 @@ public String getText() {
...
-    final float x = xy[0] + (viewWidth / 2.0f);
-    float y = xy[1] + (viewHeight / 2.0f);
+      wait.until(new Function<Void, Boolean>() {
+        @Override
+        public Boolean apply(Void input) {
+          return isLaidOut.value;
+        }
+      });
+    } catch (TimeoutException e) {
+      throw new SelendroidException("View was never laid out", e);
```
passy: I'm not quite sure if there are some standard ways of dealing with this, but it would seem useful to include some information about the view that this failed for. I guess this may also be handled on a higher level.

selendroid-server/src/main/java/io/selendroid/server/model/AndroidNativeElement.java
marcelogomez: Fair point, I'll probably move them to something like `doClick()`

selendroid-server/src/main/java/io/selendroid/server/model/AndroidNativeElement.java
marcelogomez: Since we're making this method synchronous, this will be reported as a failure in the call to the Click endpoint which should have the Selendroid element's info i.e. id o locator

marcelogomez:
@passy Any thoughts on the whole `BooleanWrapper` thing I did for syncing? Is there a better way to do this kind of waiting on a primitive value in Java

passy:
@marcelogomez No "pretty" alternatives that I'm aware of. A common pattern is

````
final boolean val[] = {false};
...
() -> { val[0] = true; }
if (val[0]) { ... }
```

passy:
Or `AtomicBoolean`.

marcelogomez:
`AtomicBoolean` looks just like the right thing :)

asm89:
@passy Thanks for the review!

PR #1138, opened 2017-11-22
Preserve original clicking behavior for API levels < 19

marcelogomez:
Fix clowniness around using CPU time instead of real time for waiting


lpetre:
Added in #1146

PR #1136, opened 2017-11-04
Virtual devices with Android SDK 25+ 

alexcomsa:
Since Android SDK version 25.3.0 the `android list avd` command has been deprecated, this affects the `DefaultAndroidEmulator.listAvailableAvds()` method  which used the previous mentioned command .
This PR adds  [avdmanager](https://developer.android.com/studio/command-line/avdmanager.html) support to obtain a list of android virtual devices when using an Android SDK version >= 25  while maintaining functionality for  the `DefaultAndroidEmulator.listAvailableAvds()` method

PR #1140, opened 2017-12-01
fix avds for avd

windysurf:
fix "avds" for "avd" small typo causing .jar file not to run.


marcelogomez:
I think this will only work for newer Android SDK installations?

lpetre:
This was fixed in #1136

PR #1143, opened 2017-12-08
Write empty args as an empty file

lpetre:
If the capabilities doesn't specific extra arguments, we get an error trying to access a missing file. So always write the extra args, even if they're empty.


lpetre:
I don't think this is necessary, actually.

PR #1144, opened 2017-12-08
Better regex matching for emulator serials, log warnings when unable to lookup a name.

lpetre:
Helps to debug issues with running selendroid on an emulator.


lpetre:
Combining this with #1145

PR #1148, opened 2017-12-14
Fix Android SDK logic for finding latest build-tools/platform version

marcelogomez:
This PR fixes the way we determine which is the latest build-tools/platform version. The previous implementation's could break if certain folders were present in an Android SDK installation, plus sorting by file name was just wrong

### DIFF
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/android/AndroidSdk.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/android/AndroidSdk.java
@@ -21,15 +21,19 @@ import io.selendroid.standalone.exceptions.AndroidSdkException;
 import java.io.File;
 import java.io.FileFilter;
 import java.util.Arrays;
-import java.util.Collections;
+import java.util.Comparator;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public class AndroidSdk {
-  private static final String ANDROID_FOLDER_PREFIX = "android-";
   private static final String ANDROID_HOME = "ANDROID_HOME";
 
   private static String sAndroidHome;
   private static String sAndroidSdkVersion;
   private static String sBuildToolsVersion;
+
+  private static final String PLATFORM_VERSION_REGEX = "android-(\\d+)$";
+  private static final String BUILD_TOOLS_VERSION_REGEX = "(\\d+)\\.(\\d+)\\.(\\d+)";
 
 
   public static File adb() {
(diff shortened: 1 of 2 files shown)


selendroid-standalone/src/main/java/io/selendroid/standalone/android/AndroidSdk.java
```diff
@@ -124,25 +108,91 @@ public static String androidJar() {
...
+
+  public static File findLatestAndroidPlatformFolder() {
+    return findLatestAndroidPlatformFolder(platformsFolder());
+  }
+
+  public static File findLatestAndroidPlatformFolder(File platformsFolder) {
+    File[] platformVersions = platformsFolder.listFiles(new FileFilter() {
+      @Override
+      public boolean accept(File file) {
+        return file.getName().matches("android-\\d+$");
```
lpetre: My inclination would be to make this regular expression a local variable, add a capture group for the integer, and then reuse that on lines 142 to get the version. Relying on a regular expression here and split below seems like it could create problems in the future if someone only updates this in one place.

selendroid-standalone/src/main/java/io/selendroid/standalone/android/AndroidSdk.java
```diff
@@ -124,25 +108,91 @@ public static String androidJar() {
...
 
-  protected static File findLatestAndroidPlatformFolder(File rootFolder, String errorMessage) {
-    File[] androidApis = rootFolder.listFiles(new AndroidFileFilter());
-    if (androidApis == null || androidApis.length == 0) {
-      throw new SelendroidException(errorMessage);
+  public static File findLatestBuildToolsFolder(File buildToolsHome) {
+    File[] buildToolsVersions = buildToolsHome.listFiles(new FileFilter() {
+      @Override
+      public boolean accept(File file) {
+        return file.getName().matches("\\d+\\.\\d+\\.\\d+$");
```
lpetre: Similar comment to above re: it would be cleaner if this regular expression had capture groups and was reused on lines 179-180 instead of splitting the string.

selendroid-standalone/src/main/java/io/selendroid/standalone/android/AndroidSdk.java
```diff
@@ -124,25 +108,91 @@ public static String androidJar() {
...
+        @Override
+        public int compare(File f1, File f2) {
+          String[] v1 = f1.getName().split("\\.");
+          String[] v2 = f2.getName().split("\\.");
+
+          for (int i = 0; i < 3; i++) {
+            int curV1 = Integer.parseInt(v1[i]);
+            int curV2 = Integer.parseInt(v2[i]);
+
+            if (curV1 != curV2) {
```
lpetre: I would call Integer.compare and then return the result if it isn't 0

PR #1151, opened 2017-12-14
Add monitoring of instrumentation process output

marcelogomez:
This PR adds monitoring of the instrumentation process (the `am instrument` call) . Whenever the instrumentation process died in the middle of a session, we wouldn't be able to detect it other than the Selendroid server on the device becoming unresponsive which lead to unnecessary timeouts. With these changes, we can check whenever the instrumentation process for the session has finished and fail immediately. This allows us to:
- Not need to timeout whenever the AUT is not installed
- Have proper native crash detection
- Get rid of `Server on the device became unreachable` messages

### DIFF
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/android/InstrumentationProcessOutput.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/android/InstrumentationProcessOutput.java
@@ -1,0 +1,136 @@
+/*
+ * Copyright 2012-2014 eBay Software Foundation and selendroid committers.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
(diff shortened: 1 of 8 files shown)


selendroid-standalone/src/main/java/io/selendroid/standalone/android/InstrumentationProcessOutput.java
```diff
@@ -0,0 +1,137 @@
...
+ package io.selendroid.standalone.android;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class InstrumentationProcessOutput {
+  private String errorMessage;
+  private String shortMessage;
+  private String longMessage;
+  private String stacktrace;
```
antiarchit: Unused field?

PR #1155, opened 2017-12-19
Immediately succeed if we find a matching serial

marcelogomez:
Serials uniquely identify device. If a serial is specified in the capabilities, we should immediately succeed or fail based on whether or not a device matches the specified serial

### DIFF
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/server/model/DeviceStore.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/server/model/DeviceStore.java
@@ -330,11 +330,17 @@ public class DeviceStore {
     return new Predicate<AndroidDevice>() {
       @Override
       public boolean apply(AndroidDevice candidate) {
+        // serial uniquely identifies a device, so we should immediately succeed
+        // if we find a match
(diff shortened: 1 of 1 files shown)


selendroid-standalone/src/main/java/io/selendroid/standalone/server/model/DeviceStore.java
```diff
@@ -330,11 +330,17 @@ public boolean apply(AndroidDevice candidate) {
     return new Predicate<AndroidDevice>() {
       @Override
       public boolean apply(AndroidDevice candidate) {
+        // serial uniquely identifies a device, so we should immediately succeed
+        // if we find a match
+        String requestedSerial = capabilities.getSerial();
+        if (StringUtils.isNotBlank(requestedSerial)) {
+          return requestedSerial.equals(candidate.getSerial());
```
alexcomsa: Maybe it would be more useful to use equalsIgnoreCase. From my perspective it's a bit more robust and less error prone.

lpetre: ADB is case sensitive, so I think the code as written is correct.

PR #1156, opened 2017-12-19
Fix app crashing detection race condition

marcelogomez:
There was a race condition between detecting regular app crashes through the instrumentation process' output. Turns out that the fastest way to detect this is through the device's crash logs.  We were also checking the crash logs in `SelendroidStandaloneDriver#waitForServerStart` which is most likely going to be executed before we even create the file.

This PR:
1. Decouples regular app crash detection from instrumentation process output checks in `ProxyToDeviceHandler`
2. Replace `while (retries) { ... sleep(...) }` with `FluentWait` in `ProxyToDeviceHandler#handleRequest`
3. Remove crash log checks in `SelendroidStandaloneDriver#waitForServerStart`
4. Makes the polling interval for waiting until the server starts configurable. Default value matches `FluentWait`'s default

### DIFF
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/SelendroidConfiguration.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/SelendroidConfiguration.java
@@ -1,11 +1,11 @@
 /*
  * Copyright 2012-2015 eBay Software Foundation and selendroid committers.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
- * 
+ *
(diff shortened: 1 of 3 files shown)


selendroid-standalone/src/main/java/io/selendroid/standalone/SelendroidConfiguration.java
```diff
@@ -161,6 +161,10 @@
       names = "-serverStartTimeout")
   private long serverStartTimeout = 20000;
 
+  @Parameter(description = "Time in milliseconds to wait between attempts to check if the selendroid-server has come up on the device.",
+      names = "-serverStartPollingInterval")
+  private long serverStartPollingInterval = 500;
```
lpetre: Can you make this explicitly use the `FluentWait` default?

PR #1158, opened 2018-01-08
Improve logging for shell command execution

marcelogomez:
This PR ads a better API for executing shell commands asynchronously and cleans up the logging of command execution output.
Instead of logging
every
single
line
separately, we now log only the final output (no more double logging for synchronous execution), along with the command that was executed. This will make logs a lot more readable.

### DIFF
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/io/ShellCommand.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/io/ShellCommand.java
@@ -34,25 +34,34 @@ public class ShellCommand {
 
   public static String exec(CommandLine commandline, long timeoutInMillies)
       throws ShellCommandException {
-    log.info("Executing shell command: " + commandline);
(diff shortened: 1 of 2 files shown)


selendroid-standalone/src/main/java/io/selendroid/standalone/io/ShellCommand.java
```diff
@@ -34,25 +34,34 @@ public static String exec(CommandLine commandLine) throws ShellCommandException
...
     }
-    String result = outputStream.getOutput().trim();
-    log.info("Shell command output\n-->\n" + result + "\n<--");
-    return result;
+
+    String output = os.toString().trim();
+    log.log(
+      Level.INFO,
+      String.format(
+        "Shell command executed: %s\n-->\n%s\n<--",
```
antiarchit: Not sure I'm a fan of the arrows, but that's how it is at the moment.

PR #1162, opened 2018-01-18
[standalone] Split local and remote port for server

marcelogomez:
Split local a remote ports used for proxying requests to the device and add a capability to enable using a random local port for forwarding

### DIFF
--- a/selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
+++ b/selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
@@ -63,13 +63,17 @@ public abstract class AbstractDevice implements AndroidDevice {
   protected String serial = null;
   protected String model = null;
   protected String apiTargetType = "android";
-  protected Integer port = null;
   protected IDevice device;
   private ByteArrayOutputStream logoutput;
   private ExecuteWatchdog logcatWatchdog;
   private static final Integer COMMAND_TIMEOUT = 20000;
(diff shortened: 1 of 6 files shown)


selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/DefaultAndroidEmulator.java
```diff
@@ -245,7 +245,7 @@ public String toString() {
   }
 
   public void setSerial(int port) {
-    this.port = port;
+    this.localPort = port;
```
lpetre: I believe this line is incorrect and should be removed

selendroid-standalone/src/main/java/io/selendroid/standalone/android/impl/AbstractDevice.java
```diff
@@ -338,14 +341,23 @@ public void forwardPort(int local, int remote) {
     }
   }
 
-  private void forwardSelendroidPort(int port) {
-    forwardPort(port, port);
+  private void forwardSelendroidPort(boolean useRandomLocalPort, int port) {
+    if (useRandomLocalPort) {
+      // Let adb forward bind to random port locally
+      String portStr = ShellCommand.exec(
+        adbCommand("forward", "tcp:0", "tcp:" + port));
```
lpetre: Can you add this behavior to `forwardPort` and return the localport from that method instead? Then you just need to optionally pass 0 to `forwardPort` as the local port and set the member variable based on the return value.

